PR #65, opened 2024-06-09
[14.0][IMP] cetmix_tower_server: 'readonly' field, many2one restrict, shared cmd/plans

norlinhenrik:
This PR has 3 commits:

**1. A record may be readonly and gray in lists**

Standard commands, flight plans, file tamplates, variables and global variable values may be managed e.g. by Odoo modules. To these 5 models, this commit adds a 'readonly' field and makes sure that 'readonly' data records are readonly and gray in list views.

**2. Restrict deleting a record referenced by many2one**

Deleting a record may delete related records only in these cases:
- Delete flight plan -> Delete related lines
- Delete flight plan line -> Delete related line actions
- Delete flight plan log -> Delete related command logs
- Delete server -> Delete related files, keys and variable values

**3. Show shared commands and plans by default**

Optional, set system parameter `cetmix_tower_server.any_server` = 1

PR #49, opened 2024-04-25
[14.0][ADD] cetmix_tower_template_odoo: add module

Aldeigja:
Cetmix Tower Odoo deployment templates

PR #66, opened 2024-06-09
[14.0][ADD] cetmix_tower_data & odoo & postgres & traefik

norlinhenrik:
Here are 4 data modules, as an alternative to `cetmix_tower_template_odoo` in #49.

The data is minimalistic. A lot of data from `cetmix_tower_template_odoo` is commented out for now.
The data is reorganized. Variable names begin with the name of the software they belong to.
The data is for standard production. A change may require a migration script (when the modules are mature).
The data is using the 'module' field introduced in #65 to be readonly and gray by default.
The data is defined in YAML. Specify that a record will not be readonly or gray by adding - before the -name.

`cetmix_tower_data` has a script to convert from YAML to XML.
`cetmix_tower_data_odoo` depends on `cetmix_tower_data_postgres` and works with `cetmix_tower_data_traefik`.


CetmixGitDrone:

ivs-cetmix:
This feature now can be implemented using `cetmix_tower_yaml` module.

PR #102, opened 2024-10-08
[14.0][IMP] cetmix_tower_server - 7 commits

norlinhenrik:
**Commit 1: A record may be readonly and gray in lists**

Standard commands, flight plans, file tamplates, variables and global variable values may be managed e.g. by Odoo modules. To these 5 models, this commit adds a 'readonly' field and makes sure that 'readonly' data records are readonly and gray in list views.

**Commit 2: Restrict deleting a record referenced by many2one**

Deleting a record may delete related records only in these cases:

Delete flight plan -> Delete related lines
Delete flight plan line -> Delete related line actions
Delete flight plan log -> Delete related command logs
Delete server -> Delete related files, keys and variable values

**Commit 3: Show shared commands and plans by default**

Optional, set system parameter cetmix_tower_server.any_server = 1

**Commit 4: Variable value may be empty.**

A server may replace a default value with an empty value.
`{% if value %}` may return `False`.

**Commit 5: Don't allow deleting a server tag if it is used by a server.**

**Commit 6: Fix something I did wrong when resolving conflicts.**

**Commit 7: Fix reference**

When searching for the existance of a variable, exclude the active record. Otherwise setting the variable reference (e.g. by a module) will result in `variable`, `variable_2`, `variable`, `variable_2 `every second time.


norlinhenrik:
Superseding #65

ivs-cetmix:
@norlinhenrik please fix pre-commit issues

norlinhenrik:
@ivs-cetmix I have fixed the pre-commit issues.

cetmix_tower_server/models/cx_tower_reference_mixin.py
```diff
@@ -66,7 +66,15 @@ def _generate_or_fix_reference(self, reference_source):
         # Check if the same reference already exists and add a suffix if yes
         counter = 1
         final_reference = reference
-        while self.search_count([("reference", "=", final_reference)]) > 0:
+        while (
```
ivs-cetmix, line 69: This method can be called when no record exists yet. Eg [here](https://github.com/cetmix/cetmix-tower/blob/50c240f9c4b628043fddaf76ee5e4d84c30866a1/cetmix_tower_server/models/cx_tower_reference_mixin.py#L122)

So we need to check id only in case if it's called for record. Something like:
```python
domain = [("reference", "=", final_reference)]
if self:
    domain.append(("id", "not in", self.ids)),
```

We are using `self.ids` here bacause this method can be also called for a recordset. Eg in the `write` method [here](https://github.com/cetmix/cetmix-tower/blob/50c240f9c4b628043fddaf76ee5e4d84c30866a1/cetmix_tower_server/models/cx_tower_reference_mixin.py#L157)

ivs-cetmix (changes requested):
@norlinhenrik thank you for your contribution!
Let's follow "one PR - one feature" policy in order to simplify review and merge process.
Please mark this PR as draft and leave it as a reference.
Let's start with the [commit](https://github.com/cetmix/cetmix-tower/pull/102/commits/495382be8e27434fa64a8996323898d7036f70ed) that fixes reference generation. Please move it int into a new branch and open a new PR.
You should also check my comments regarding it. And please don't forget to add/modify tests when doing some code modifications.
I would even suggest to start with tests, ensure it fails and than updating the code for the test to pass.

CetmixGitDrone:

cetmix_tower_server/models/cx_tower_reference_mixin.py
ivs-cetmix: Hey @norlinhenrik could you please check #108 ?
It features a fix for this issue.

norlinhenrik:
@ivs-cetmix 
So commit 7 you have implemented already 👍 , and commit 6 is related to commit 2.

Before I write tests and make pull requests, I would like to know if there are any of the 5 first commits which you think should not be implemented?

ivs-cetmix:
Hi @norlinhenrik please check the latest updates that were done with https://github.com/cetmix/cetmix-tower/pull/126
Look like it closes some of your pending issues.

ivs-cetmix:
Hey @norlinhenrik  please check https://github.com/cetmix/cetmix-tower/pull/132 
It should solve the issue from Commit 4

ivs-cetmix:
Finally solved with https://github.com/cetmix/cetmix-tower/pull/151
@norlinhenrik thank you for your contribution once again!

PR #112, opened 2024-10-15
[14.0][IMP] cetmix_tower_yaml Many2one fields support

ivs-cetmix:
Add support for many2one fields export/import
-----------------------------------------------

Field value is saved in yaml as a related record reference.
Eg 'file_template_id: file_template_1' where 'file_template_1' is the 'reference' field value of the related record.

Post fix update
---------------
Enable previously disabled test that was skipped due to an error in reference generation (t3982).

### DIFF
--- a/cetmix_tower_yaml/models/cx_tower_yaml_mixin.py
+++ b/cetmix_tower_yaml/models/cx_tower_yaml_mixin.py
@@ -6,6 +6,7 @@ import yaml
 
 from odoo import _, fields, models
 from odoo.exceptions import ValidationError
+from odoo.tools import ormcache
 
 
 class CustomDumper(yaml.Dumper):
@@ -42,6 +43,10 @@ class CxTowerYamlMixin(models.AbstractModel):
     )
     yaml_file = fields.Binary(compute="_compute_yaml_code", attachment=False)
     yaml_file_name = fields.Char(compute="_compute_yaml_code")
+    yaml_explode = fields.Boolean(
+        string="Explode",
+        help="Add entire related record data instead of just a reference",
+    )
 
     def _compute_yaml_code(self):
         """Compute YAML code based on model record data"""
@@ -52,10 +57,8 @@ class CxTowerYamlMixin(models.AbstractModel):
         for record in self:
             # We are reading field list for each record
             # because list of fields can differ from record to record
-            yaml_keys = record._get_fields_for_yaml()
-            record_dict = record.read(fields=yaml_keys)[0]
             yaml_code = yaml.dump(
-                record._post_process_record_values(record_dict),
+                record._prepare_record_for_yaml(),
                 Dumper=CustomDumper,
                 default_flow_style=False,
             )
@@ -74,6 +77,17 @@ class CxTowerYamlMixin(models.AbstractModel):
                 record_yaml_dict = yaml.safe_load(record.yaml_code)
                 record_vals = record._post_process_yaml_dict_values(record_yaml_dict)
                 record.update(record_vals)
+
+    def _prepare_record_for_yaml(self):
+        """Reads and processes current record before converting it to YAML
+
+        Returns:
+            dict: values ready for YAML conversion
+        """
+        self.ensure_one()
+        yaml_keys = self._get_fields_for_yaml()
+        record_dict = self.read(fields=yaml_keys)[0]
+        return self._post_process_record_values(record_dict)
 
     def _get_fields_for_yaml(self):
         """Get ist of field to be present in YAML
@@ -110,6 +124,24 @@ class CxTowerYamlMixin(models.AbstractModel):
             values.update(
                 {"access_level": self.TO_YAML_ACCESS_LEVEL[values["access_level"]]}
             )
+
+        # Check if we need to return a record dict or just a reference
+        # Use context value first, revert to the record setting if not defined
+        explode_related_record = self._context.get("explode_related_record")
+        if explode_related_record is None:
+            explode_related_record = self.yaml_explode
+
+        # Post process m2o fields
+        for key, value in values.items():
+            # IMPORTANT: Odoo naming patterns must be followed for related fields.
+            # This is why we are checking for the field name ending here.
+            # Further checks for the field type are done in _process_m2o_value()
+            if key.endswith("_id"):
+                processed_value = self.with_context(
+                    explode_related_record=explode_related_record
+                )._process_m2o_value(key, value, record_mode=True)
+                values.update({key: processed_value})
+
         return values
 
     def _post_process_yaml_dict_values(self, values):
@@ -123,7 +155,7 @@ class CxTowerYamlMixin(models.AbstractModel):
         """
 
         # Check Cetmix Tower YAML version
-        yaml_version = values.pop("cetmix_tower_yaml_version")
+        yaml_version = values.pop("cetmix_tower_yaml_version", None)
         if (
             yaml_version
             and isinstance(yaml_version, int)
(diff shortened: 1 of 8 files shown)


cetmix_tower_yaml/models/cx_tower_yaml_mixin.py
```diff
@@ -110,8 +120,59 @@ def _post_process_record_values(self, values):
             values.update(
                 {"access_level": self.TO_YAML_ACCESS_LEVEL[values["access_level"]]}
             )
+        # Post process m2o fields
+        for key, value in values.items():
+            if key.endswith("_id"):
```
GabbasovDinar, line 139: in this case there is no guarantee that the specified field will be exactly the relation field, it is necessary to check exactly its type without focusing on how the field is called:

```python

model_fields = self._fields
for field_name, field_value in values.items():
    field = model_fields.get(field_name)
    if field and field["type"] == "many2one":
        values.update({
            field_name: self._process_m2o_value(field, field_value, record_mode=True)
        })

```

cetmix_tower_yaml/models/cx_tower_yaml_mixin.py
```diff
@@ -110,8 +120,59 @@ def _post_process_record_values(self, values):
...
+                    # Record -> Yaml
+                    if record_mode:
+                        record = self.env[comodel_name].browse(value[0])
+                        return record and record.reference
+
+                    # Yaml -> Record
+                    else:
+                        record = self.env[comodel_name].get_by_reference(value)
+                        return record and record.id
+
```
GabbasovDinar:
```suggestion
        if not value:
            return False
        
        comodel_name = field.comodel_name
        
        # Process value
        if comodel_name and self._model_supports_yaml(comodel_name):
            # Record -> Yaml
            if record_mode:
                record = self.env[comodel_name].browse(value[0])
                return record and record.reference
        
            # Yaml -> Record
            else:
                record = self.env[comodel_name].get_by_reference(value)
                return record and record.id
        else:
            return False

```

cetmix_tower_yaml/models/cx_tower_yaml_mixin.py
```diff
@@ -110,8 +120,59 @@ def _post_process_record_values(self, values):
             values.update(
                 {"access_level": self.TO_YAML_ACCESS_LEVEL[values["access_level"]]}
             )
+        # Post process m2o fields
+        for key, value in values.items():
+            if key.endswith("_id"):
+                processed_value = self._process_m2o_value(key, value, record_mode=True)
+                values.update({key: processed_value})
```
GabbasovDinar, line 143: what if processed_value returned None, because the model is not support yaml? from this eval `if comodel_name and self._model_supports_yaml(comodel_name):`

GabbasovDinar (changes requested):

cetmix_tower_yaml/models/cx_tower_yaml_mixin.py
ivs-cetmix: @GabbasovDinar I agree. However we can imply naming conventions at the project level.
It's also needed because of the YAML-> Odoo conversion. There simply no field types there so the only thing we can rely on is the field suffix.

cetmix_tower_yaml/models/cx_tower_yaml_mixin.py
ivs-cetmix: Then None value will be assigned to the key. Which is fine

cetmix_tower_yaml/models/cx_tower_yaml_mixin.py
ivs-cetmix: @GabbasovDinar it's also checked [here](https://github.com/cetmix/cetmix-tower/blob/996f9a6f4e541a30b6831d5418783328398f7517/cetmix_tower_yaml/models/cx_tower_yaml_mixin.py#L151)

cetmix_tower_yaml/models/cx_tower_yaml_mixin.py
GabbasovDinar: @ivs-cetmix in this case I suggest to return False explicitly so that there is no confusion in reading the code

cetmix_tower_yaml/models/cx_tower_yaml_mixin.py
GabbasovDinar: @ivs-cetmix Why not add field type validation as I pointed out above?

GabbasovDinar: In this case there is no need to remember about any YAML file, and it is immediately clear from the code that a specific type of field is expected. the code is also cleaner and clearer

cetmix_tower_yaml/models/cx_tower_yaml_mixin.py
GabbasovDinar: @ivs-cetmix large nested code structure

cetmix_tower_yaml/models/cx_tower_yaml_mixin.py
```diff
@@ -161,4 +193,125 @@ def _post_process_yaml_dict_values(self, values):
...
+                    )
+                # No reference and value is string
+                else:
+                    record = False
+
+                # Assign final result
+                result = record and record.id
+
+            # Return
+            return result
```
GabbasovDinar: Some refactoring to improve readability and simplicity

```python
def _process_m2o_value(self, field, value, record_mode=False):
    """Post process many2one value
        Args:
            field (Char): Field the value belongs to
            value (Char): Value to process
            record_mode (Bool): If True process value as a record value
                                else process value as a YAML value
            Context:
                explode_related_record: if set will return entire record dictionary
                    not just a reference
        Returns:
            dict() or Char: record dictionary if fetch_record else reference
    """
    
    # Step 1: Return False if the value is not set or the field is not found
    if not value or not (field_obj := self._fields.get(field)):
        return False

    # Step 2: Return False if the field is not a many2one field or has no comodel
    if field_obj.type != "many2one" or not field_obj.comodel_name:
        return False

    comodel = self.env[field_obj.comodel_name]
    explode_related_record = self._context.get("explode_related_record")

    # Step 3: Process value when working in record mode (Record -> YAML)
    if record_mode:
        # Retrieve the record based on the ID provided in the value
        record = comodel.browse(value[0])

        # If the context specifies to explode the related record, return its dictionary representation
        if explode_related_record:
            return record._prepare_record_for_yaml() if record else False

        # Otherwise, return just the reference (or False if record does not exist)
        return record.reference if record else False

    # Step 4: Process value in normal mode (YAML -> Record)
    
    # If the value is a string, it is treated as a reference
    if isinstance(value, str):
        reference = value

    # If the value is a dictionary, extract the reference from it
    elif isinstance(value, dict):
        reference = value.get("reference")

        # If reference is found, retrieve the corresponding record
        if reference:
            record = comodel.get_by_reference(reference)

            # If the record exists, update it with the values from the dictionary
            if record:
                record.write(record._post_process_yaml_dict_values(value))
            # If the record does not exist, create a new one
            else:
                record = comodel.create(comodel._post_process_yaml_dict_values(value))

        # If there's no reference but value is a dict, create a new record
        else:
            record = comodel.create(comodel._post_process_yaml_dict_values(value))

        # Return the record's ID if it exists, otherwise return False
        return record.id if record else False

    # If the value is neither string nor dict, return False
    else:
        return False

    # Step 5: Final fallback: attempt to retrieve the record by reference if set, return its ID or False
    record = comodel.get_by_reference(reference) if reference else False
    return record.id if record else False



```

GabbasovDinar (changes requested):

cetmix_tower_yaml/models/cx_tower_yaml_mixin.py
ivs-cetmix: @GabbasovDinar thank you very much for such a detailed and well documented suggestion 👍

GabbasovDinar (approved):

GabbasovDinar:
LGTM!

CetmixGitDrone:

Aldeigja (approved):
functional LGTM

CetmixGitDrone:
Prepared branch [14.0-dev-ocabot-merge-pr-112-by-ivs-cetmix-bump-nobump](https://github.com/cetmix/cetmix-tower/commits/14.0-dev-ocabot-merge-pr-112-by-ivs-cetmix-bump-nobump), awaiting test results.

CetmixGitDrone:
Congratulations, your PR was merged at f98bdc05dd17eb8d60136aea5365e42a8fcc995d. Thanks a lot for contributing to cetmix. ❤️

PR #113, opened 2024-10-15
[14.0][FIX] cetmix_tower_server: Clean up Python command execution result

dessanhemrayev:
Renamed method _parse_ssh_command_results to  _parse_command_results for using not only for SSH but for any command including Python one.

Task: 3952

### DIFF
--- a/cetmix_tower_server/models/cx_tower_key.py
+++ b/cetmix_tower_server/models/cx_tower_key.py
@@ -450,6 +450,9 @@ class CxTowerKey(models.Model):
 
         # Replace keys with values
         for key_value in key_values:
+            # If key_value includes quotes, remove them for the replacement
+            if key_value.startswith('"') and key_value.endswith('"'):
+                key_value = key_value[1:-1]
             # Replace key including key terminator
             code = code.replace(key_value, self.SECRET_VALUE_SPOILER)
 
--- a/cetmix_tower_server/models/cx_tower_key.py
+++ b/cetmix_tower_server/models/cx_tower_key.py
@@ -451,8 +451,7 @@ class CxTowerKey(models.Model):
         # Replace keys with values
         for key_value in key_values:
             # If key_value includes quotes, remove them for the replacement
-            if key_value.startswith('"') and key_value.endswith('"'):
-                key_value = key_value[1:-1]
+            key_value = key_value.strip('"')
             # Replace key including key terminator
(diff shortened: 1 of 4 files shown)


cetmix_tower_server/tests/test_command.py
```diff
@@ -837,3 +844,27 @@ def test_execute_python_code(self):
             command_result["error"],
             "Error in command result must be set to None",
         )
+
+    def test_execute_python_code_with_secret(self):
```
ivs-cetmix, line 869: We need to check if secret key is replaced in the command output. I didn't manage to find such a flow in this test.

cetmix_tower_server/tests/test_command.py
```diff
@@ -41,6 +41,19 @@ def setUp(self, *args, **kwargs):
     COMMAND_RESULT = {"exit_code": 0, "message": "New command was created"}
 else:
     COMMAND_RESULT = {"exit_code": -1, "message": "error"}
+    """,
+            }
+        )
+
+        self.command_create_new_command_2 = self.Command.create(
+            {
+                "name": "Create new command",
```
ivs-cetmix: Command name as well as the name of the variable it's assigned don't look relevant. Which degraded the code readability.
There should be something like "Python command with secret" or similar.

ivs-cetmix (changes requested):

cetmix_tower_server/tests/test_command.py
dessanhemrayev: Thanks, I'll fix it

cetmix_tower_server/tests/test_command.py
dessanhemrayev: Thanks, I'll fix it

cetmix_tower_server/models/cx_tower_key.py
```diff
@@ -450,6 +450,9 @@ def _replace_with_spoiler(self, code, key_values):
 
         # Replace keys with values
         for key_value in key_values:
+            # If key_value includes quotes, remove them for the replacement
+            if key_value.startswith('"') and key_value.endswith('"'):
+                key_value = key_value[1:-1]
```
GabbasovDinar:
```suggestion
            key_value = key_value.strip('"')
```

GabbasovDinar (changes requested):

PR #114, opened 2024-10-20
[14.0][IMP] t3995 cetmix_tower_server: references in tags

dessanhemrayev:
This commit adds references to tags for easier export/import and action referencing.

Changes
- Inherit from cx.tower.reference.mixin
- Remove field
- Add test
- Add reference fields to tag views, including search view
- Update documentation to reflect new reference fields in tags

Task: 3995

### DIFF
--- a/cetmix_tower_server/README.rst
+++ b/cetmix_tower_server/README.rst
@@ -17,7 +17,7 @@ Cetmix Tower Server Management
     :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
     :alt: License: AGPL-3
 .. |badge3| image:: https://img.shields.io/badge/github-cetmix%2Fcetmix--tower-lightgray.png?logo=github
-    :target: https://github.com/cetmix/cetmix-tower/tree/14.0-dev/cetmix_tower_server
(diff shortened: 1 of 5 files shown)


cetmix_tower_server/tests/test_reference_mixin.py
```diff
@@ -92,3 +92,16 @@ def test_search_by_reference(self):
         # Search using malformed (case sensitive)
         search_result = self.ServerTemplate.get_by_reference("not_much_template")
         self.assertEqual(len(search_result), 0, "Result should be empty")
+
+    def test_search_tag_by_reference(self):
```
ivs-cetmix: This function is inherited from mixin and is already tested there. We are not modifying it, so no need to add tests for it too.

ivs-cetmix (changes requested):
Code is good, just no need for tests here)

cetmix_tower_server/tests/test_reference_mixin.py
dessanhemrayev: Thanks, I'll remove it

CetmixGitDrone:

PR #129, opened 2024-10-29
[14.0][IMP] cetmix_tower_server: add reference to models

dessanhemrayev:
This commit adds references to the following models:

- cx.tower.plan.line
- cx.tower.plan.line.action
- cx.tower.variable.value

Changes

Inherit from cx.tower.reference.mixin
Add reference fields to views, including search view 
Update tests

Task: 4017

## Release Notes

- **New Features**
  - Enhanced duplication process for flight plans and variable values, ensuring integrity during copying.
  - New `create` methods added to various models to manage references effectively during record creation.
  - New `reference` field added to the `cx.tower.plan.line`, `cx.tower.variable.value`, and `cx.tower.plan` views, improving data handling and visibility control based on user permissions.

- **Bug Fixes**
  - Improved test cases for plan execution and command handling, ensuring accurate validation.

- **Tests**
  - Expanded test coverage for reference preparation and population functionalities.

### DIFF
--- a/cetmix_tower_server/models/cx_tower_plan.py
+++ b/cetmix_tower_server/models/cx_tower_plan.py
@@ -294,13 +294,15 @@ class CxTowerPlan(models.Model):
 
             # Duplicate actions linked to the line
             for action in line.action_ids:
-                new_action = action.copy(
+                new_action = action.with_context(reference_mixin_skip_copy=True).copy(
                     # link new actions to the new line
-                    {"line_id": new_line.id}
+                    {"line_id": new_line.id, "name": line.name}
                 )
 
                 # Duplicate variable values linked to the action
                 for variable_value in action.variable_value_ids:
-                    variable_value.copy({"plan_line_action_id": new_action.id})
+                    variable_value.with_context(reference_mixin_skip_self=True).copy(
+                        {"plan_line_action_id": new_action.id}
+                    )
 
         return new_plan
--- a/cetmix_tower_server/models/cx_tower_plan_line.py
+++ b/cetmix_tower_server/models/cx_tower_plan_line.py
@@ -10,6 +10,9 @@ from .constants import PLAN_LINE_CONDITION_CHECK_FAILED
 
 class CxTowerPlanLine(models.Model):
     _name = "cx.tower.plan.line"
+    _inherit = [
+        "cx.tower.reference.mixin",
+    ]
     _order = "sequence, plan_id"
     _description = "Cetmix Tower Flight Plan Line"
 
@@ -193,3 +196,11 @@ class CxTowerPlanLine(models.Model):
             is_skipped=True,
             **log_vals,
         )
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        # Use the common method from mixin
+        vals_list = self._populate_references(
+            "cx.tower.plan", "plan_id", vals_list, suffix="_line"
+        )
+        return super().create(vals_list)
--- a/cetmix_tower_server/models/cx_tower_plan_line_action.py
+++ b/cetmix_tower_server/models/cx_tower_plan_line_action.py
@@ -1,10 +1,11 @@
 # Copyright (C) 2022 Cetmix OÜ
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
 from odoo import _, api, fields, models
 
 
 class CxTowerPlanLineAction(models.Model):
-    _inherit = ["cx.tower.variable.mixin"]
+    _inherit = ["cx.tower.variable.mixin", "cx.tower.reference.mixin"]
     _name = "cx.tower.plan.line.action"
     _description = "Cetmix Tower Flight Plan Line Action"
 
@@ -76,3 +77,11 @@ class CxTowerPlanLineAction(models.Model):
                 )
             else:
                 rec.name = _("Wrong action")
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        # Use the common method from mixin with a custom suffix
+        vals_list = self._populate_references(
+            "cx.tower.plan.line", "line_id", vals_list, suffix="_action"
+        )
+        return super().create(vals_list)
--- a/cetmix_tower_server/models/cx_tower_reference_mixin.py
+++ b/cetmix_tower_server/models/cx_tower_reference_mixin.py
@@ -1,6 +1,7 @@
 # Copyright (C) 2024 Cetmix OÜ
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 import re
+from collections import defaultdict
 
 from odoo import _, api, fields, models
 from odoo.osv import expression
@@ -68,13 +69,11 @@ class CxTowerReferenceMixin(models.AbstractModel):
         final_reference = reference
 
         # If exclude same records from search results
-        if self:
+        if self and not self.env.context.get("reference_mixin_skip_self"):
             domain = [("id", "not in", self.ids)]
         else:
             domain = []
-
         final_domain = expression.AND([domain, [("reference", "=", final_reference)]])
-
         while self.search_count(final_domain) > 0:
             counter += 1
             final_reference = _(f"{reference}_{counter}")
@@ -129,7 +128,8 @@ class CxTowerReferenceMixin(models.AbstractModel):
         """
         if not self._context.get("reference_mixin_override"):
             for vals in vals_list:
-                vals["name"] = vals["name"].strip()
+                if vals.get("name"):
+                    vals["name"] = vals["name"].strip()
                 # Generate reference
                 reference = self._generate_or_fix_reference(
                     vals.get("reference") or vals.get("name")
@@ -186,10 +186,14 @@ class CxTowerReferenceMixin(models.AbstractModel):
         copy_name = _("%(name)s (copy)", name=original_name)
 
         counter = 1
+        # Ensures that the generated copy name is unique by
+        # appending a counter until a unique name is found.
         while self.search_count([("name", "=", copy_name)]) > 0:
             counter += 1
             copy_name = _(
-                "%(name)s (copy %(number)s)", name=original_name, number=str(counter)
+                "%(name)s (copy %(number)s)",
+                name=original_name,
+                number=str(counter),
             )
 
         return copy_name
@@ -208,7 +212,11 @@ class CxTowerReferenceMixin(models.AbstractModel):
         self.ensure_one()
         if default is None:
             default = {}
-        default["name"] = self._get_copied_name()
+
+        # skip copy name because this function use in models
+        # where it field name non store
+        if not self.env.context.get("reference_mixin_skip_copy"):
+            default["name"] = self._get_copied_name()
         if "reference" not in default:
             default["reference"] = self._generate_or_fix_reference(default["name"])
         return super().copy(default=default)
@@ -242,3 +250,93 @@ class CxTowerReferenceMixin(models.AbstractModel):
 
         # This is in case some models will remove reference uniqueness constraint
         return records and records[0].id
+
+    @api.model
+    def _prepare_references(self, model, key_name, vals_list):
+        """
+        Prepare a dictionary of references for given model records.
+
+        This function extracts unique IDs from a list of dictionaries (vals_list)
+        based on a specified key (key_name), fetches the corresponding records
+        from the specified model, and returns a dictionary mapping record IDs to
+        their references.
+
+        Args:
+            model (str): The name of the model to fetch records from.
+            key_name (str): The key in the dictionaries of vals_list that contains
+                            the record IDs.
+            vals_list (list of dict): A list of dictionaries containing the values
+                                    to be processed.
+
+        Returns:
+            dict: A dictionary mapping record IDs to their references.
+        """
+        if not vals_list:
+            # No entries to process, return an empty dictionary
+            return {}
+
+        try:
+            CxModel = self.env[model]
+        except KeyError as err:
+            raise ValueError(
+                _(
+                    (
+                        "Model '%(model)s' does not exist. "
+                        "Please provide a valid model name."
+                    ),
+                    model=model,
+                )
+            ) from err
+
+        # Extract all unique ids from vals_list
+        line_ids = {vals.get(key_name) for vals in vals_list if vals.get(key_name)}
+
+        # Fetch all line references in a single query
+        lines = CxModel.browse(line_ids)
+        return {line.id: line.reference for line in lines}
+
+    @api.model
+    def _populate_references(self, model_name, field_name, vals_list, suffix=""):
+        """
+        Populates reference fields in a list of dictionaries (vals_list)
+        intended for record creation.
+
+        This method generates unique references for each dictionary entry in
+        `vals_list` based on a specified field that links to records in
+        another model (indicated by `model_name`). It uses existing references
+        from the related records as a basis and appends a suffix and an
+        incrementing index to ensure uniqueness.
+
+        Args:
+            model_name (str): The name of the related model to extract
+                              reference data from.
+            field_name (str): The key in each dictionary in `vals_list`
+                              containing the related record's ID.
+            vals_list (list of dict): A list of dictionaries where each dictionary
+                               represents values for a new record.
+            suffix (str, optional): A suffix to append to each generated reference.
+                               Defaults to an empty string.
+
+        Returns:
+            list: The modified `vals_list`, with a unique 'reference'
+                  entry in each dictionary.
+        """
+
+        # Extract unique references from vals_list
+        refs = self._prepare_references(model_name, field_name, vals_list)
+        line_index_dict = defaultdict(int)
+
+        # Populate vals with references
+        for vals in vals_list:
+            record_id = vals.get(field_name)
+            if record_id and refs.get(record_id):
+                line_index_dict[record_id] += 1
+                line_index = line_index_dict[record_id]
+                vals["reference"] = f"{refs[record_id]}{suffix}_{line_index}"
+            else:
+                # Handle cases where the field is not present
+                line_index_dict["no_record"] += 1
+                line_index = line_index_dict["no_record"]
+                vals["reference"] = f"no_{suffix}_{line_index}"
+
+        return vals_list
--- a/cetmix_tower_server/models/cx_tower_variable_value.py
+++ b/cetmix_tower_server/models/cx_tower_variable_value.py
@@ -1,6 +1,7 @@
 # Copyright (C) 2022 Cetmix OÜ
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
+
 from odoo import _, api, fields, models
 from odoo.exceptions import ValidationError
 from odoo.osv.expression import OR
@@ -9,12 +10,16 @@ from odoo.osv.expression import OR
 class TowerVariableValue(models.Model):
     _name = "cx.tower.variable.value"
     _description = "Cetmix Tower Variable Values"
+    _inherit = [
+        "cx.tower.reference.mixin",
+    ]
     _rec_name = "variable_reference"
     _order = "variable_reference"
 
     variable_id = fields.Many2one(
         string="Variable", comodel_name="cx.tower.variable", required=True
     )
+    name = fields.Char(related="variable_id.name", readonly=True)
     variable_reference = fields.Char(
         related="variable_id.reference", store=True, index=True
     )
@@ -276,3 +281,11 @@ class TowerVariableValue(models.Model):
                         var=record.variable_id.name,
                     )
                 )
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        # Use the common method from mixin
+        vals_list = self._populate_references(
+            "cx.tower.variable", "variable_id", vals_list, suffix="_variable"
+        )
+        return super().create(vals_list)
--- a/cetmix_tower_server/tests/test_reference_mixin.py
+++ b/cetmix_tower_server/tests/test_reference_mixin.py
@@ -8,6 +8,21 @@ class TestTowerReference(TestTowerCommon):
     We are using ServerTemplate for that.
     """
 
+    def setUp(self, *args, **kwargs):
+        super().setUp(*args, **kwargs)
+
+        self.plan_reference_mixin = self.Plan.create(
+            {"name": "Test Plan reference mixin", "note": "Test Note reference mixin"}
+        )
+
+        self.plan_line_reference_mixin = self.plan_line.create(
+            {
+                "plan_id": self.plan_reference_mixin.id,
+                "sequence": 1,
+                "command_id": self.command_list_dir.id,
+            }
+        )
+
     def test_reference_generation(self):
         """Test reference generation"""
 
@@ -92,3 +107,122 @@ class TestTowerReference(TestTowerCommon):
         # Search using malformed (case sensitive)
         search_result = self.ServerTemplate.get_by_reference("not_much_template")
         self.assertEqual(len(search_result), 0, "Result should be empty")
+
+    def test_prepare_references_valid_input(self):
+        """
+        Ensure references are correctly prepared for valid input.
+        """
+
+        vals_list = [{"plan_id": self.plan_reference_mixin.id}]
+        result = self.plan_line._prepare_references(
+            "cx.tower.plan", "plan_id", vals_list
+        )
+
+        # Verify the result contains the expected reference
+        self.assertIn(
+            self.plan_reference_mixin.id,
+            result,
+            "The reference ID should be in the result.",
+        )
+        self.assertEqual(
+            result[self.plan_reference_mixin.id],
+            self.plan_reference_mixin.reference,
+            "The reference should match the expected value.",
+        )
+
+    def test_prepare_references_invalid_model_name(self):
+        """
+        Check that an error is raised for an invalid model name.
+        """
+
+        vals_list = [{"plan_id": self.plan_reference_mixin.id}]
+        with self.assertRaises(ValueError) as cm:
+            self.plan_line._prepare_references("invalid.model", "plan_id", vals_list)
+
+        # Confirm the exception message is as expected
+        self.assertEqual(
+            str(cm.exception),
+            "Model 'invalid.model' does not exist. Please provide a valid model name.",
+            "The error message should indicate an invalid model name.",
+        )
+
+    def test_prepare_references_empty_vals_list(self):
+        """
+        Verify that an empty vals_list returns an empty dictionary.
+        """
+        result = self.plan_line._prepare_references("cx.tower.plan", "plan_id", [])
+        self.assertEqual(
+            result,
+            {},
+            "The result should be an empty dictionary when vals_list is empty.",
+        )
+
+    def test_populate_references_with_valid_input(self):
+        """
+        Ensure references are populated correctly in the provided values list.
+        """
+        vals_list = [{"plan_id": self.plan_reference_mixin.id}]
+        updated_vals = self.plan_line._populate_references(
+            "cx.tower.plan", "plan_id", vals_list, suffix="TEST"
+        )
+
+        # Check the updated values contain the expected reference format
+        self.assertEqual(
+            updated_vals[0]["reference"],
+            f"{self.plan_reference_mixin.reference}TEST_1",
+            "The reference should be correctly populated with the suffix.",
+        )
+
+    def test_populate_references_missing_field(self):
+        """
+        Confirm that entries missing the required field are handled properly.
+        """
+
+        vals_list_with_missing_field = [{"another_key": 123}]
+        updated_vals_with_missing = self.plan_line._populate_references(
+            "cx.tower.plan", "plan_id", vals_list_with_missing_field, suffix="MISSING"
+        )
+        self.assertEqual(
+            updated_vals_with_missing[0]["reference"],
+            "no_MISSING_1",
+            "Entries missing the required field should have a default reference.",
+        )
+
+    def test_populate_references_duplicate_ids(self):
+        """
+        Ensure that duplicate IDs in the input list are correctly
+        handled and referenced.
+        """
+        vals_list = [
+            {"plan_id": self.plan_reference_mixin.id},
+            {"plan_id": self.plan_reference_mixin.id},
+        ]
+        updated_vals = self.plan_line._populate_references(
+            "cx.tower.plan", "plan_id", vals_list, suffix="DUPLICATE"
+        )
+
+        # Verify that each duplicate entry has a unique suffix
+        self.assertEqual(
+            updated_vals[0]["reference"],
+            f"{self.plan_reference_mixin.reference}DUPLICATE_1",
+            "The first duplicate reference should have the correct suffix.",
+        )
+        self.assertEqual(
+            updated_vals[1]["reference"],
+            f"{self.plan_reference_mixin.reference}DUPLICATE_2",
+            "The second duplicate reference should have the correct suffix.",
+        )
+
+    def test_populate_references_empty_vals_list(self):
+        """
+        Check that an empty input list returns an empty result
+        when populating references.
+        """
+        updated_vals = self.plan_line._populate_references(
+            "cx.tower.plan", "plan_id", [], suffix="EMPTY"
+        )
+        self.assertEqual(
+            updated_vals,
+            [],
+            "The result should be an empty list when vals_list is empty.",
+        )
--- a/cetmix_tower_server/views/cx_tower_plan_line_view.xml
+++ b/cetmix_tower_server/views/cx_tower_plan_line_view.xml
@@ -10,7 +10,10 @@
                     <group>
                         <group>
                             <field name="sequence" />
-
+                            <field
+                                name="reference"
+                                placeholder="Can contain English letters, digits and '_'. Leave blank to autogenerate"
+                            />
                             <field
                                 name="command_id"
                                 context="{'command_show_server_names': True}"
@@ -38,6 +41,7 @@
                                 <tree>
                                     <field name="sequence" widget="handle" />
                                     <field name="name" />
+                                    <field name="reference" optional="hide" />
                                 </tree>
                                 <form>
                                     <group>
@@ -65,8 +69,13 @@
                                         >
                                             <tree editable="bottom">
                                                 <field
+                                                    name="reference"
+                                                    optional="hide"
+                                                />
+                                                <field
                                                     name="variable_reference"
-                                                    invisible="1"
+                                                    optional="hide"
+                                                    string="Variable Reference"
                                                 />
                                                 <field name="variable_id" />
                                                 <field name="value_char" />
--- a/cetmix_tower_server/views/cx_tower_variable_value_view.xml
+++ b/cetmix_tower_server/views/cx_tower_variable_value_view.xml
@@ -1,11 +1,11 @@
 <?xml version="1.0" encoding="utf-8" ?>
 <odoo>
-
     <record id="cx_tower_variable_value_view_tree" model="ir.ui.view">
         <field name="name">cx.tower.variable.value.view.tree</field>
         <field name="model">cx.tower.variable.value</field>
         <field name="arch" type="xml">
             <tree editable="top" decoration-bf="is_global==True">
+                <field name="reference" optional="hidden" />
                 <field name="variable_id" />
                 <field name="is_global" widget="boolean_toggle" />
                 <field
@@ -21,12 +21,12 @@
             </tree>
         </field>
     </record>
-
     <record id="cx_tower_variable_value_search_view" model="ir.ui.view">
         <field name="name">cx.tower.variable.value.view.search</field>
         <field name="model">cx.tower.variable.value</field>
         <field name="arch" type="xml">
             <search string="Search Values">
+                <field name="reference" />
                 <field name="variable_id" string="Variable" />
                 <field name="value_char" string="Value" />
                 <filter
@@ -39,15 +39,13 @@
                     name="global"
                     domain="[('is_global', '=', True)]"
                 />
-             </search>
+            </search>
         </field>
     </record>
-
     <record id="action_cx_tower_variable_value" model="ir.actions.act_window">
         <field name="name">Variable Values</field>
         <field name="type">ir.actions.act_window</field>
         <field name="res_model">cx.tower.variable.value</field>
         <field name="view_mode">tree</field>
     </record>
-
 </odoo>
--- a/cetmix_tower_server/models/cx_tower_server_template.py
+++ b/cetmix_tower_server/models/cx_tower_server_template.py
@@ -358,7 +358,9 @@ class CxTowerServerTemplate(models.Model):
 
         # Duplicate variable values
         for variable_value in self.variable_value_ids:
-            variable_value.copy({"server_template_id": new_template.id})
+            variable_value.with_context(reference_mixin_skip_self=True).copy(
+                {"server_template_id": new_template.id}
+            )
 
         # Duplicate server logs
         for server_log in self.server_log_ids:
--- a/cetmix_tower_server/tests/test_plan.py
+++ b/cetmix_tower_server/tests/test_plan.py
@@ -367,18 +367,18 @@ class TestTowerPlan(TestTowerCommon):
         command_1 = self.Command.create(
             {"name": "New Test Command", "access_level": "3"}
         )
-        self.plan_line_2_1 = self.plan_line.create(
-            {
-                "sequence": 5,
-                "command_id": command_1.id,
-            }
-        )
 
         self.plan_2 = self.Plan.create(
             {
                 "name": "Test plan 2",
                 "note": "Create directory and list its content",
-                "line_ids": [(4, self.plan_line_2_1.id)],
+            }
+        )
+        self.plan_line_2_1 = self.plan_line.create(
+            {
+                "sequence": 5,
+                "plan_id": self.plan_2.id,
+                "command_id": command_1.id,
             }
         )
         self.assertTrue(self.plan_2.access_level_warn_msg)
@@ -663,12 +663,12 @@ class TestTowerPlan(TestTowerCommon):
                 "command_id": command.id,
                 "path": "/test/path",
                 # Condition based on Linux version
-                "condition": '{{ linux_version }} >= "5.0"',
+                "condition": '{{ test_linux_version }} >= "5.0"',
             }
         )
 
         # Create a variable for the action
-        variable = self.Variable.create({"name": "linux_version"})
+        variable = self.Variable.create({"name": "test_linux_version"})
 
         # Create an Action for the Plan Line
         action = self.plan_line_action.create(
--- a/cetmix_tower_server/views/cx_tower_plan_view.xml
+++ b/cetmix_tower_server/views/cx_tower_plan_view.xml
@@ -69,6 +69,7 @@
                                 <tree decoration-bf="action=='plan'">
                                     <field name="sequence" widget="handle" />
                                     <field name="name" />
+                                    <field name="reference" optional="hide" />
                                     <field
                                         name="action"
                                         optional="show"


cetmix_tower_server/views/cx_tower_variable_value_view.xml
```diff
@@ -1,11 +1,12 @@
 <?xml version="1.0" encoding="utf-8" ?>
 <odoo>
-
     <record id="cx_tower_variable_value_view_tree" model="ir.ui.view">
         <field name="name">cx.tower.variable.value.view.tree</field>
         <field name="model">cx.tower.variable.value</field>
         <field name="arch" type="xml">
             <tree editable="top" decoration-bf="is_global==True">
+                <field name="name" />
```
ivs-cetmix: This field has no sense for this model. So no need to show it

cetmix_tower_server/views/cx_tower_variable_value_view.xml
```diff
@@ -1,11 +1,12 @@
...
+                <field name="reference" optional="show" />
```
ivs-cetmix: Should be hidden by default

cetmix_tower_server/views/cx_tower_variable_value_view.xml
```diff
@@ -21,12 +22,13 @@
             </tree>
         </field>
     </record>
-
     <record id="cx_tower_variable_value_search_view" model="ir.ui.view">
         <field name="name">cx.tower.variable.value.view.search</field>
         <field name="model">cx.tower.variable.value</field>
         <field name="arch" type="xml">
             <search string="Search Values">
+                <field name="name" />
```
ivs-cetmix: This field doesn't have any use for this model, remove it.

cetmix_tower_server/views/cx_tower_server_view.xml
```diff
@@ -237,6 +237,7 @@
                             <field name="variable_value_ids" order="variable_reference">
                                 <tree editable="bottom">
                                     <field name="variable_reference" invisible="1" />
+                                    <field name="name" />
```
ivs-cetmix: Remove this field, no-one will be creating names manually

cetmix_tower_server/views/cx_tower_plan_line_view.xml
```diff
@@ -10,7 +10,11 @@
                     <group>
                         <group>
                             <field name="sequence" />
-
+                            <field name="name" />
```
ivs-cetmix: This field is technical in this model, no need to show it

cetmix_tower_server/models/cx_tower_plan_line.py
```diff
@@ -10,12 +10,15 @@
...
+    _inherit = [
+        "cx.tower.reference.mixin",
+    ]
     _order = "sequence, plan_id"
     _description = "Cetmix Tower Flight Plan Line"
 
     active = fields.Boolean(related="plan_id.active", readonly=True)
     sequence = fields.Integer(default=10)
-    name = fields.Char(related="command_id.name", readonly=True)
+    name = fields.Char(related="command_id.name", store=True)
```
ivs-cetmix: Probably we can try to leave it as is (non-stored, readonly) and see if it works.

cetmix_tower_server/models/cx_tower_plan.py
```diff
@@ -182,6 +182,7 @@ def _get_next_action_values(self, command_log):
                                 "variable_id": variable.id,
                                 "value_char": variable_value.value_char,
                                 "server_id": server.id,
+                                "name": variable_value.value_char,
```
ivs-cetmix: Why do we need it? And why `variable_value.value_char` is used for the "name" field value?

cetmix_tower_server/models/cetmix_tower.py
```diff
@@ -61,6 +61,7 @@ def server_set_variable_value(self, server_reference, variable_reference, value)
         else:
             self.env["cx.tower.variable.value"].create(
                 {
+                    "name": value,
```
ivs-cetmix: This field should not be defined manually

ivs-cetmix (changes requested):
It's not clear how and why someone should provide "name" values for technical records, such as plan lines, plan line actions and variable values.
Basically "name" field value should be either autogenerated (bad) or used from related model. We need to check if this is possible to make "name" field non-stored in order to avoid useless data storage.

cetmix_tower_server/models/cx_tower_plan.py
dessanhemrayev: Thanks, I'll fix it

cetmix_tower_server/models/cx_tower_plan_line.py
```diff
@@ -10,6 +12,9 @@
 
 class CxTowerPlanLine(models.Model):
     _name = "cx.tower.plan.line"
+    _inherit = [
+        "cx.tower.reference.mixin",
```
GabbasovDinar, line 14: Since this mixin has been added you also need to change the `_compute_variable_ids` method. please check how to use this method in the case when a mixin is connected: e.g. [here](https://github.com/cetmix/cetmix-tower/blob/fd29f987a08f3eb9ae3f4ce902f1cc969a6f082f/cetmix_tower_server/models/cx_tower_command.py#L123)

cetmix_tower_server/models/cx_tower_plan_line.py
```diff
@@ -187,3 +192,10 @@ def _skip(self, server, plan_log_record, **kwargs):
             is_skipped=True,
             **log_vals,
         )
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        for vals in vals_list:
+            current_datetime = datetime.now().strftime("%Y-%m-%d %H:%M:%S.%f")
```
GabbasovDinar: Please move this before the cycle starts

cetmix_tower_server/models/cx_tower_variable_value.py
```diff
@@ -9,10 +11,14 @@
 class TowerVariableValue(models.Model):
     _name = "cx.tower.variable.value"
     _description = "Cetmix Tower Variable Values"
+    _inherit = [
+        "cx.tower.reference.mixin",
```
GabbasovDinar, line 14: please, fix `_compute_variable_ids` method

cetmix_tower_server/models/cx_tower_plan_line.py
GabbasovDinar: you can use datetime from field:
```python
fields.Datetime.now()
```

Also use Server datetime format (I recommend using this method):
```python
from odoo.tools.misc import DEFAULT_SERVER_DATETIME_FORMAT
```

If this format does not suit you, then define your format as a constant in constants.py: 
```python
PLAN_LINE_DATETIME_FORMAT = "%Y-%m-%d %H:%M:%S.%f"
```

cetmix_tower_server/models/cx_tower_plan_line_action.py
```diff
@@ -60,19 +59,53 @@ def _compute_name(self):
...
+    @api.model_create_multi
+    def create(self, vals_list):
+        for vals in vals_list:
+            name = self._generate_action_name(
+                vals.get("condition"),
+                vals.get("action"),
+                vals.get("value_char"),
+                vals.get("custom_exit_code"),
+            )
+            vals.update({"name": name if name else "Plan Line Action"})
```
GabbasovDinar: @ivs-cetmix Should be a translatable value this default value ? ("Plan Line Action")

cetmix_tower_server/models/cx_tower_reference_mixin.py
```diff
@@ -186,11 +186,14 @@ def _get_copied_name(self):
...
-                "%(name)s (copy %(number)s)", name=original_name, number=str(counter)
-            )
+        if not self.env.context.get("skip_copy_action"):
+            while self.search_count([("name", "=", copy_name)]) > 0:
+                counter += 1
+                copy_name = _(
+                    "%(name)s (copy %(number)s)",
+                    name=original_name,
+                    number=str(counter),
+                )
```
GabbasovDinar: you can reduce code nesting, which will improve its readability:

```python
if self.env.context.get("skip_copy_action"):
    return copy_name

while .......
```

please, check it.

And please add comments as to why we should use this context and only then execute the loop

cetmix_tower_server/models/cx_tower_variable_value.py
```diff
@@ -277,3 +283,10 @@ def _check_single_assignment(self):
                         "Server, Server Template, or Plan Line Action."
                     )
                 )
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        for vals in vals_list:
+            current_datetime = datetime.now().strftime("%Y-%m-%d %H:%M:%S.%f")
```
GabbasovDinar: use fields.Datetime.now() and see my comments above

GabbasovDinar (changes requested):

cetmix_tower_server/models/cx_tower_plan_line.py
```diff
@@ -187,3 +192,10 @@ def _skip(self, server, plan_log_record, **kwargs):
...
+            vals["name"] = f"Plan line {current_datetime}"
```
ivs-cetmix: Why do we need to create a name for a plan line? It's not used anywhere

cetmix_tower_server/models/cx_tower_plan_line_action.py
```diff
@@ -60,19 +59,53 @@ def _compute_name(self):
...
+        """
+        action_selection_vals = dict(self._fields["action"].selection)  # type: ignore
+
+        # Compose name based on values
+        if condition and action and value_char:
+            action_string = action_selection_vals.get(action)
+
+            # Add custom exit code if action presumes it
+            if action == "ec":
+                action_string = "{} {}".format(action_string, custom_exit_code)
```
ivs-cetmix: Please use f-string instead of 'format'

cetmix_tower_server/models/cx_tower_reference_mixin.py
```diff
@@ -186,11 +186,14 @@ def _get_copied_name(self):
         copy_name = _("%(name)s (copy)", name=original_name)
 
         counter = 1
-        while self.search_count([("name", "=", copy_name)]) > 0:
-            counter += 1
-            copy_name = _(
-                "%(name)s (copy %(number)s)", name=original_name, number=str(counter)
-            )
+        if not self.env.context.get("skip_copy_action"):
```
ivs-cetmix: Why do we need it?

cetmix_tower_server/models/cx_tower_variable_value.py
```diff
@@ -277,3 +283,10 @@ def _check_single_assignment(self):
...
+            vals["name"] = f"Variable value {current_datetime}"
```
ivs-cetmix: Why do we need it? Are we using field 'name' somewhere?

ivs-cetmix (changes requested):

cetmix_tower_server/models/cx_tower_reference_mixin.py
dessanhemrayev: Because our "name" field is not saved, a problem arises when copying "plan line action". The essence of the problem is that the [method ](https://github.com/cetmix/cetmix-tower/blob/fd29f987a08f3eb9ae3f4ce902f1cc969a6f082f/cetmix_tower_server/models/cx_tower_reference_mixin.py#L190)that generates the value for the "reference" field

cetmix_tower_server/models/cx_tower_plan_line.py
dessanhemrayev: You are right, but the generation method depends on this field. To avoid possible problems, we are forced to add the value this way. Also, the reference must be unique, so we include the date in them.

cetmix_tower_server/models/cx_tower_plan_line.py
dessanhemrayev: > Please move this before the cycle starts

I need unique data to create a reference (the "name" field is used to create a reference)

cetmix_tower_server/models/cx_tower_plan_line.py
dessanhemrayev: Thanks, I'll check it and fix

cetmix_tower_server/models/cx_tower_variable_value.py
dessanhemrayev: I tested this method, because it rounds the values ​​I can't use it
```
datetime.datetime.now()
2024-10-31 19:54:14.310060

fields.Datetime.now() 
2024-10-31 19:54:14.000000
```

cetmix_tower_server/models/cx_tower_plan_line.py
ivs-cetmix: > > Please move this before the cycle starts
> 
> I need unique data to create a reference (the "name" field is used to create a reference)

@dessanhemrayev  'name' field should'n be unique for an unique reference to be generated. This case is already handled in the [create](https://github.com/cetmix/cetmix-tower/blob/2711a9e613a0c3f7728cf176665e78f74ec8372a/cetmix_tower_server/models/cx_tower_reference_mixin.py#L137) method of the mixin.

cetmix_tower_server/models/cx_tower_plan_line.py
ivs-cetmix: I agree, however it's [related field](https://github.com/cetmix/cetmix-tower/blob/2711a9e613a0c3f7728cf176665e78f74ec8372a/cetmix_tower_server/models/cx_tower_plan_line.py#L22) 
Meanwhile [reference](https://github.com/cetmix/cetmix-tower/blob/2711a9e613a0c3f7728cf176665e78f74ec8372a/cetmix_tower_server/models/cx_tower_reference_mixin.py#L138) field is also used during the creation of the record. 
So I would suggest something like this:

```python
@api.model_create_multi
def create(self, vals_list):
     PlanModel = self.env["cx.tower.plan"]

     # Store the last line index per plan
     plan_line_index_dict = {}

     # Populate vals with references
     for vals in vals_list:
        plan_id = vals.get("plan_id")
        # Get stored line index for this plan and increment it
        line_index = plan_line_index_dict.get(plan_id, 1) + 1
        # Line reference = <plan_reference>_<line_index>
        vals.update({"reference": f"{PlanModel.browse(plan_id).reference}_{line_index}"})
        # Save updated line index
        plan_line_index_dict.update({plan_id: line_index})

    # Let main function do the job
    return super().create(vals_list)
```

@GabbasovDinar your thoughts on that?

cetmix_tower_server/models/cx_tower_plan_line_action.py
```diff
@@ -60,19 +59,53 @@ def _compute_name(self):
...
-                        action_string,
-                    )
+                rec.name = self._generate_action_name(
+                    rec.condition, rec.action, rec.value_char, rec.custom_exit_code
                 )
             else:
                 rec.name = _("Wrong action")
+
+    @api.model
+    def _generate_action_name(self, condition, action, value_char, custom_exit_code):
```
ivs-cetmix: Why do we need to make this a separate function?

cetmix_tower_server/models/cx_tower_plan_line_action.py
ivs-cetmix: @

> @ivs-cetmix Should be a translatable value this default value ? ("Plan Line Action")

@GabbasovDinar nope

cetmix_tower_server/models/cx_tower_plan_line_action.py
```diff
@@ -60,19 +59,53 @@ def _compute_name(self):
...
+                    value_char,
+                    _("then"),
+                    action_string,
+                )
+            )
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        for vals in vals_list:
+            name = self._generate_action_name(
```
ivs-cetmix: Same as with plan lines - we can generate 'referecne' instead of 'name'

cetmix_tower_server/models/cx_tower_variable_value.py
ivs-cetmix: @GabbasovDinar what's wrong with it?

cetmix_tower_server/models/cx_tower_plan_line.py
ivs-cetmix: @

> Since this mixin has been added you also need to change the `_compute_variable_ids` method. please check how to use this method in the case when a mixin is connected: e.g. [here](https://github.com/cetmix/cetmix-tower/blob/fd29f987a08f3eb9ae3f4ce902f1cc969a6f082f/cetmix_tower_server/models/cx_tower_command.py#L123)

@GabbasovDinar Could you please elaborate why this should be done?

cetmix_tower_server/models/cx_tower_reference_mixin.py
ivs-cetmix: > Because our "name" field is not saved, a problem arises when copying "plan line action". The essence of the problem is that the [method ](https://github.com/cetmix/cetmix-tower/blob/fd29f987a08f3eb9ae3f4ce902f1cc969a6f082f/cetmix_tower_server/models/cx_tower_reference_mixin.py#L190)that generates the value for the "reference" field

In this case you should use more relevant name, because 'skip_copy' is too generic.
Eg 'reference_mixin_skip_copy'

ivs-cetmix (changes requested):

cetmix_tower_server/models/cx_tower_plan_line.py
GabbasovDinar: @ivs-cetmix we need to refactor this method because now this model inherits the methods of mixin

cetmix_tower_server/models/cx_tower_variable_value.py
GabbasovDinar: @ivs-cetmix you need to compare these methods when the mixin is connected and when it is not. method uses the methods of the mixin, in one case by accessing the mixin model (because it is not connected) and in the second case using inherited methods, so, we need refactoring this method, because the mixin was connected with the model.

cetmix_tower_server/models/cx_tower_variable_value.py
GabbasovDinar: why? add comments to code, I don't understand

cetmix_tower_server/models/cx_tower_plan_line.py
GabbasovDinar: if I understand the problem correctly, then

```python

from collections import defaultdict


@api.model_create_multi
def create(self, vals_list):
    PlanModel = self.env["cx.tower.plan"]

    # Extract all unique plan_ids from vals_list
    plan_ids = {vals.get("plan_id") for vals in vals_list if vals.get("plan_id")}

    # Fetch all plan references in a single query
    plans = PlanModel.browse(plan_ids)
    plan_refs = {plan.id: plan.reference for plan in plans}

    # Initialize line indices for each plan_id
    plan_line_index_dict = defaultdict(int)

    # Populate vals with references
    for vals in vals_list:
        plan_id = vals.get("plan_id")
        plan_line_index_dict[plan_id] += 1
        line_index = plan_line_index_dict[plan_id]
        vals["reference"] = f"{plan_refs[plan_id]}_{line_index}"

    # Delegate creation to the superclass method
    return super().create(vals_list)
```

GabbasovDinar (changes requested):
please, check comments

cetmix_tower_server/models/cx_tower_plan_line.py
dessanhemrayev: @ivs-cetmix  @GabbasovDinar 
what about if plan_id not in vals ?
eg: [here](https://github.com/cetmix/cetmix-tower/blob/2711a9e613a0c3f7728cf176665e78f74ec8372a/cetmix_tower_server/tests/test_plan.py#L37)

dessanhemrayev: and what about this [line](https://github.com/cetmix/cetmix-tower/blob/2711a9e613a0c3f7728cf176665e78f74ec8372a/cetmix_tower_server/models/cx_tower_reference_mixin.py#L135)?

cetmix_tower_server/models/cx_tower_reference_mixin.py
dessanhemrayev: Thanks, I'll fix it

cetmix_tower_server/models/cx_tower_plan_line.py
ivs-cetmix: > what about if plan_id not in vals ? eg: [here](https://github.com/cetmix/cetmix-tower/blob/2711a9e613a0c3f7728cf176665e78f74ec8372a/cetmix_tower_server/tests/test_plan.py#L37)

@dessanhemrayev thank you for pointing on this. In this case we can create some default placeholder reference. Eg "line_with_no_plan"

ivs-cetmix: > and what about this [line](https://github.com/cetmix/cetmix-tower/blob/2711a9e613a0c3f7728cf176665e78f74ec8372a/cetmix_tower_server/models/cx_tower_reference_mixin.py#L135)?

@dessanhemrayev good point! I think we need to add a condition here. Eg `if "name" in vals: ...`

cetmix_tower_server/models/cx_tower_plan_line.py
```diff
@@ -187,3 +191,32 @@ def _skip(self, server, plan_log_record, **kwargs):
...
+        # Populate vals with references
+        for vals in vals_list:
+            plan_id = vals.get("plan_id")
+            if plan_id:
+                plan_line_index_dict[plan_id] += 1
+                line_index = plan_line_index_dict[plan_id]
+                vals["reference"] = f"{plan_refs[plan_id]}_{line_index}"
+            else:
+                # Handle the case where plan_id is not present
+                tmp_line_index = self._get_current_microsecond()
```
ivs-cetmix: We can simply use some static value instead. It will be automatically modified by [this](https://github.com/cetmix/cetmix-tower/blob/9e8617fcdb7aa7c1c5cbbdbf193d9e5caf208ee8/cetmix_tower_server/models/cx_tower_reference_mixin.py#L42) function if needed.
You are using similar method in variable values, so why are you using milliseconds based approach here?

cetmix_tower_server/models/cx_tower_plan_line_action.py
ivs-cetmix: @dessanhemrayev this is still pending

cetmix_tower_server/models/cx_tower_plan_line_action.py
ivs-cetmix: @dessanhemrayev this is still pending

cetmix_tower_server/models/cx_tower_variable_value.py
ivs-cetmix: @GabbasovDinar are you sure that you are talking about the reference mixin here?

cetmix_tower_server/models/cx_tower_reference_mixin.py
```diff
@@ -186,10 +188,17 @@ def _get_copied_name(self):
         copy_name = _("%(name)s (copy)", name=original_name)
 
         counter = 1
+        if self.env.context.get("reference_mixin_skip_copy"):
```
ivs-cetmix: You are skipping copy and returning copied name at the same time. Why?

ivs-cetmix (changes requested):

cetmix_tower_server/models/cx_tower_plan_line.py
dessanhemrayev: Thanks
You are right, but when we run the tests, [this function](https://github.com/cetmix/cetmix-tower/blob/9e8617fcdb7aa7c1c5cbbdbf193d9e5caf208ee8/cetmix_tower_server/models/cx_tower_reference_mixin.py#L42) does not work correctly
yes, it checks for the presence of such a value in the table and shows that it is unique, but still, when creating a record, the sql_constraints is triggered and the creation of the record is blocked

and because of this I had to use such a function

cetmix_tower_server/models/cx_tower_plan_line_action.py
dessanhemrayev: I'll remove this function

cetmix_tower_server/models/cx_tower_reference_mixin.py
dessanhemrayev: I'll fix it

cetmix_tower_server/models/cx_tower_plan_line.py
ivs-cetmix: @dessanhemrayev this means that either that function or tests should be fixed. Using milliseconds creates barely usable references and besides from the theory it doesn't give a 100% guarantee from the duplicates.

cetmix_tower_server/models/cx_tower_plan_line.py
```diff
@@ -187,3 +191,32 @@ def _skip(self, server, plan_log_record, **kwargs):
             is_skipped=True,
             **log_vals,
         )
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        PlanModel = self.env["cx.tower.plan"]
+
+        # Extract all unique plan_ids from vals_list
+        plan_ids = {vals.get("plan_id") for vals in vals_list if vals.get("plan_id")}
```
ivs-cetmix: Although sets a faster for search creating and updating them is slower than lists.  Here is some [Stackoverflow](https://stackoverflow.com/questions/7717011/which-is-faster-and-why-set-or-list) for that.
Probably `add x to list if x is not in list` will be faster. I would suggest profiling it to get the answer.

cetmix_tower_server/models/cx_tower_plan_line_action.py
```diff
@@ -76,3 +78,33 @@ def _compute_name(self):
...
+        for vals in vals_list:
+            line_id = vals.get("line_id")
+            if line_id:
+                plan_line_index_dict[line_id] += 1
+                line_index = plan_line_index_dict[line_id]
+                vals["reference"] = f"{plan_refs[line_id]}_action_{line_index}"
+            else:
+                # Handle the case where line_id is not present
+                plan_line_index_dict["no_line"] += 1
+                line_index = plan_line_index_dict["no_line"]
```
ivs-cetmix: Where is that "no_line" key added in dict?

cetmix_tower_server/models/cx_tower_reference_mixin.py
```diff
@@ -208,7 +214,14 @@ def copy(self, default=None):
         self.ensure_one()
         if default is None:
             default = {}
-        default["name"] = self._get_copied_name()
+
+        # skip copy name because this function use in models
+        # where it field name non store
+        default["name"] = (
```
ivs-cetmix: copy() function will use that original name automatically. No need to set it in code.

cetmix_tower_server/models/cx_tower_variable_value.py
```diff
@@ -277,3 +284,34 @@ def _check_single_assignment(self):
                         "Server, Server Template, or Plan Line Action."
                     )
                 )
+
+    @api.model_create_multi
+    def create(self, vals_list):
```
ivs-cetmix, line 286: Same issues as in all other `create` function implementations.

ivs-cetmix (changes requested):

cetmix_tower_server/models/cx_tower_plan_line.py
dessanhemrayev: Thanks
I checked it link and tested code block
```python
import random
import timeit

# Generate a list of dictionaries with random values
vals_list = [{'key_name': random.randint(0, 1000)} for _ in range(10000)]
key_name = 'key_name'

def func1():
    line_ids = []
    for vals in vals_list:
        line_id = vals.get(key_name)
        if line_id and line_id not in line_ids:
            line_ids.append(line_id)

def func2():
    plan_ids = {vals.get(key_name) for vals in vals_list if vals.get(key_name)}

# Benchmark the two functions
time_func1 = timeit.timeit(func1, number=100)
time_func2 = timeit.timeit(func2, number=100)

print(f"Time taken by func1: {time_func1:.6f} seconds")
print(f"Time taken by func2: {time_func2:.6f} seconds")

```
I got these results
image

cetmix_tower_server/models/cx_tower_plan_line_action.py
dessanhemrayev: if you are talking raise a KeyError
we use defaultdict and defaultdict never raises a KeyError

cetmix_tower_server/models/cx_tower_variable_value.py
```diff
@@ -277,3 +284,29 @@ def _check_single_assignment(self):
...
+
+                vals["reference"] = f"{variable_refs[variable_id]}_{line_index}"
+            else:
+                # Handle the case where variable_id is not present
+                variable_line_index_dict["no_variable"] += 1
+                line_index = variable_line_index_dict["no_variable"]
+                vals["reference"] = f"value_with_no_variable_{line_index}"
+
+        # Delegate creation to the superclass method
+        return super().create(vals_list)
```
GabbasovDinar, line 291: this method has the same code and is used in three different places, please make one common method (maybe in cx.tower.reference.mixin model) and use it, it will make it easier to maintain the code in the future

GabbasovDinar (changes requested):

cetmix_tower_server/models/cx_tower_variable_value.py
dessanhemrayev: Thanks, I'll fix it

cetmix_tower_server/models/cx_tower_reference_mixin.py
```diff
@@ -242,3 +250,66 @@ def _get_id_by_reference(self, reference):
 
         # This is in case some models will remove reference uniqueness constraint
         return records and records[0].id
+
+    @api.model
+    def _prepare_references(self, model, key_name, vals_list):
```
ivs-cetmix, line 255: This function must be covered with tests

cetmix_tower_server/models/cx_tower_reference_mixin.py
```diff
@@ -242,3 +250,66 @@ def _get_id_by_reference(self, reference):
...
+
+        # Extract all unique ids from vals_list
+        line_ids = {vals.get(key_name) for vals in vals_list if vals.get(key_name)}
+
+        # Fetch all line references in a single query
+        lines = CxModel.browse(line_ids)
+        return {line.id: line.reference for line in lines}
+
+    @api.model
+    def _populate_references(self, model_name, field_name, vals_list, suffix=""):
```
ivs-cetmix, line 299: This function must be covered with tests

cetmix_tower_server/models/cx_tower_reference_mixin.py
```diff
@@ -68,13 +69,11 @@ def _generate_or_fix_reference(self, reference_source):
         final_reference = reference
 
         # If exclude same records from search results
-        if self:
+        if self and not self.env.context.get("reference_mixin_skip_self"):
```
ivs-cetmix, line 72: Why do we need to skip it?

cetmix_tower_server/models/cx_tower_reference_mixin.py
```diff
@@ -129,7 +128,8 @@ def create(self, vals_list):
         """
         if not self._context.get("reference_mixin_override"):
             for vals in vals_list:
-                vals["name"] = vals["name"].strip()
+                if "name" in vals:
+                    vals["name"] = vals["name"].strip()
```
ivs-cetmix, line 132: What if `name == False?`

ivs-cetmix (changes requested):
You should cover your code with tests right when you write it. Especially for such complex scenarios where TDD (test driven development) is preferred.

cetmix_tower_server/models/cx_tower_reference_mixin.py
dessanhemrayev: when we copy the value of a variable, we cannot verify the uniqueness, because this point is added to the domain of another condition

cetmix_tower_server/models/cx_tower_reference_mixin.py
```diff
@@ -242,3 +250,71 @@ def _get_id_by_reference(self, reference):
...
+                                    to be processed.
+
+        Returns:
+            dict: A dictionary mapping record IDs to their references.
+        """
+        try:
+            CxModel = self.env[model]
+        except KeyError as err:
+            raise ValueError(
+                f"Model '{model}' does not exist. Please provide a valid model name."
```
ivs-cetmix: Messages should be translated properly

cetmix_tower_server/tests/common.py
```diff
@@ -208,6 +208,18 @@ def setUp(self, *args, **kwargs):
             }
         )
 
+        self.plan_reference_mixin = self.Plan.create(
```
ivs-cetmix: This method is used in one test only, so it should be moved to that test `setUp`

cetmix_tower_server/tests/common.py
```diff
@@ -208,6 +208,18 @@ def setUp(self, *args, **kwargs):
...
+            {"name": "Test Plan reference mixin", "note": "Test Note reference mixin"}
+        )
+
+        self.plan_line_reference_mixin = self.plan_line.create(
```
ivs-cetmix: Same here

cetmix_tower_server/models/cx_tower_reference_mixin.py
ivs-cetmix: Sorry, don't understand this quite well. Could you please elaborate it further with some examples?

cetmix_tower_server/models/cx_tower_reference_mixin.py
```diff
@@ -242,3 +250,71 @@ def _get_id_by_reference(self, reference):
...
+        line_ids = {vals.get(key_name) for vals in vals_list if vals.get(key_name)}
+
+        # Fetch all line references in a single query
+        lines = CxModel.browse(line_ids)
+        return {line.id: line.reference for line in lines}
+
+    @api.model
+    def _populate_references(self, model_name, field_name, vals_list, suffix=""):
+        """
+        Common method to populate references based on a specific field.
```
ivs-cetmix: I still cannot understand what does this function do. 
Description should be more specific. Eg `Create record references in the vals list used for 'create' method. Use references of the related model as a basis`.  Or something like that.

cetmix_tower_server/models/cx_tower_reference_mixin.py
```diff
@@ -242,3 +250,71 @@ def _get_id_by_reference(self, reference):
...
+
+        Args:
+            model_name (str): The model name to extract references from.
```
ivs-cetmix: Maybe 'Related model name'? Or can this be not a related record either?

cetmix_tower_server/tests/test_reference_mixin.py
```diff
@@ -92,3 +92,109 @@ def test_search_by_reference(self):
         # Search using malformed (case sensitive)
         search_result = self.ServerTemplate.get_by_reference("not_much_template")
         self.assertEqual(len(search_result), 0, "Result should be empty")
+
+    def test_prepare_references_valid_input(self):
```
ivs-cetmix, line 111: Which function exactly is tested here? Opt for using `test_<function_name>_..` naming pattern for better readability.

cetmix_tower_server/tests/test_reference_mixin.py
```diff
@@ -92,3 +92,109 @@ def test_search_by_reference(self):
...
+            result,
+            "The reference ID should be in the result.",
+        )
+        self.assertEqual(
+            result[self.plan_reference_mixin.id],
+            self.plan_reference_mixin.reference,
+            "The reference should match the expected value.",
+        )
+
+    def test_invalid_model_name_raises_error(self):
```
ivs-cetmix: Same regarding the test name here

cetmix_tower_server/tests/test_reference_mixin.py
```diff
@@ -92,3 +92,109 @@ def test_search_by_reference(self):
...
+            self.plan_line._prepare_references("invalid.model", "plan_id", vals_list)
+
+        # Confirm the exception message is as expected
+        self.assertEqual(
+            str(cm.exception),
+            "Model 'invalid.model' does not exist. Please provide a valid model name.",
+            "The error message should indicate an invalid model name.",
+        )
+
+    def test_empty_vals_list_returns_empty_dict(self):
```
ivs-cetmix: And here

cetmix_tower_server/tests/test_reference_mixin.py
```diff
@@ -92,3 +92,109 @@ def test_search_by_reference(self):
...
+        """Verify that an empty vals_list returns an empty dictionary."""
+        result = self.plan_line._prepare_references("cx.tower.plan", "plan_id", [])
+        self.assertEqual(
+            result,
+            {},
+            "The result should be an empty dictionary when vals_list is empty.",
+        )
+
+    def test_references_populated_correctly(self):
```
ivs-cetmix: And here

cetmix_tower_server/tests/test_reference_mixin.py
```diff
@@ -92,3 +92,109 @@ def test_search_by_reference(self):
...
+        )
+
+        # Check the updated values contain the expected reference format
+        self.assertEqual(
+            updated_vals[0]["reference"],
+            f"{self.plan_reference_mixin.reference}TEST_1",
+            "The reference should be correctly populated with the suffix.",
+        )
+
+    def test_handling_missing_field_in_vals_list(self):
```
ivs-cetmix: And here

cetmix_tower_server/tests/test_reference_mixin.py
```diff
@@ -92,3 +92,109 @@ def test_search_by_reference(self):
...
+        updated_vals_with_missing = self.plan_line._populate_references(
+            "cx.tower.plan", "plan_id", vals_list_with_missing_field, suffix="MISSING"
+        )
+        self.assertEqual(
+            updated_vals_with_missing[0]["reference"],
+            "no_MISSING_1",
+            "Entries missing the required field should have a default reference.",
+        )
+
+    def test_handling_duplicate_ids(self):
```
ivs-cetmix: And here

cetmix_tower_server/tests/test_reference_mixin.py
```diff
@@ -92,3 +92,109 @@ def test_search_by_reference(self):
...
+            f"{self.plan_reference_mixin.reference}DUPLICATE_1",
+            "The first duplicate reference should have the correct suffix.",
+        )
+        self.assertEqual(
+            updated_vals[1]["reference"],
+            f"{self.plan_reference_mixin.reference}DUPLICATE_2",
+            "The second duplicate reference should have the correct suffix.",
+        )
+
+    def test_empty_vals_list_for_populate_references(self):
```
ivs-cetmix: And here

ivs-cetmix (changes requested):

cetmix_tower_server/tests/test_reference_mixin.py
dessanhemrayev: Thanks, I'll fix it

cetmix_tower_server/models/cx_tower_reference_mixin.py
dessanhemrayev: Thanks, I'll fix it

cetmix_tower_server/models/cx_tower_reference_mixin.py
dessanhemrayev: okay,I'll fix it

cetmix_tower_server/models/cx_tower_reference_mixin.py
dessanhemrayev: Thanks, I'll fix it

cetmix_tower_server/models/cx_tower_reference_mixin.py
dessanhemrayev: when we copy
```python
variable_value.copy({"plan_line_action_id": new_action.id})
```
we can't check its reference for uniqueness
because here (line 72) in self (old_record)
we got domain where 
`domain = [("id", "not in", self.ids), ("reference","=", reference)] `
and it sometimes incorrect

cetmix_tower_server/models/cx_tower_reference_mixin.py
```diff
@@ -129,7 +128,8 @@
         """
         if not self._context.get("reference_mixin_override"):
             for vals in vals_list:
-                vals["name"] = vals["name"].strip()
+                if vals.get("name"):
+                    vals["name"] = vals["name"].strip()
```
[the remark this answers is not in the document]
dessanhemrayev: Thank you for pointing this out! After consideration, I realized that calling strip() on an empty string has no effect—it will remain empty. Therefore, we don't need to modify the condition to handle empty strings explicitly. The current condition if vals.get("name"): is sufficient, as it ensures strip() is only called when name has a meaningful, non-empty value.

This approach also skips any unnecessary processing on empty or None values, improving efficiency.

cetmix_tower_server/models/cx_tower_reference_mixin.py
```diff
@@ -242,3 +250,84 @@ def _get_id_by_reference(self, reference):
...
+        Args:
+            model (str): The name of the model to fetch records from.
+            key_name (str): The key in the dictionaries of vals_list that contains
+                            the record IDs.
+            vals_list (list of dict): A list of dictionaries containing the values
+                                    to be processed.
+
+        Returns:
+            dict: A dictionary mapping record IDs to their references.
+        """
```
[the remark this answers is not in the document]
dessanhemrayev: Thank you for pointing this out! I'll fix it

cetmix_tower_server/models/cx_tower_reference_mixin.py
```diff
@@ -242,3 +250,84 @@
...
+        refs = self._prepare_references(model_name, field_name, vals_list)
+        line_index_dict = defaultdict(int)
+
+        # Populate vals with references
+        for vals in vals_list:
+            record_id = vals.get(field_name)
+            if record_id:
+                line_index_dict[record_id] += 1
+                line_index = line_index_dict[record_id]
+                vals["reference"] = f"{refs[record_id]}{suffix}_{line_index}"
```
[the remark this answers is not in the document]
dessanhemrayev: Thank you for pointing this out! I'll fix it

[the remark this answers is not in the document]
dessanhemrayev: Thank you for pointing this out! I'll fix it

GabbasovDinar (approved):

cetmix_tower_server/models/cx_tower_reference_mixin.py
```diff
@@ -242,3 +250,88 @@ def _get_id_by_reference(self, reference):
...
+        """
+        if not vals_list:
+            # No entries to process, return an empty dictionary
+            return {}
+
+        try:
+            CxModel = self.env[model]
+        except KeyError as err:
+            raise ValueError(
+                _("Model '%s' does not exist. Please provide a valid model name.")
```
ivs-cetmix: Please use correct translated strings implementation

CetmixGitDrone:

cetmix_tower_server/models/cx_tower_plan_line.py
```diff
@@ -10,6 +10,9 @@
 
 class CxTowerPlanLine(models.Model):
     _name = "cx.tower.plan.line"
+    _inherit = [
+        "cx.tower.reference.mixin",
+    ]
```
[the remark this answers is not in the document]
ivs-cetmix: This is not relevant

cetmix_tower_server/models/cx_tower_reference_mixin.py
ivs-cetmix: What does it mean "incorrect"?

cetmix_tower_server/models/cx_tower_variable_value.py
```diff
@@ -9,12 +10,14 @@
...
+    _inherit = [
+        "cx.tower.reference.mixin",
+    ]
     _rec_name = "variable_reference"
     _order = "variable_reference"
 
-    variable_id = fields.Many2one(
-        string="Variable", comodel_name="cx.tower.variable", required=True
-    )
+    variable_id = fields.Many2one(string="Variable", comodel_name="cx.tower.variable")
```
ivs-cetmix: Why do you remove "required" attribute here?

cetmix_tower_server/models/cx_tower_variable_value.py
```diff
@@ -279,3 +282,11 @@ def _check_single_assignment(self):
...
                     )
                 )
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        # Use the common method from mixin
+        vals_list = self._populate_references(
+            "cx.tower.variable", "variable_id", vals_list, suffix="_variable"
+        )
+        return super().create(vals_list)
```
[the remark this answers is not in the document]
ivs-cetmix: This method is used only in several model.

ivs-cetmix (changes requested):

cetmix_tower_server/models/cx_tower_reference_mixin.py
dessanhemrayev: Maybe I expressed myself incorrectly. What I meant was that if the search domain is like this, we will not be able to get a reliable result because the domain limits the choice of data.

cetmix_tower_server/models/cx_tower_reference_mixin.py
```diff
@@ -242,3 +250,90 @@
...
+                    model=model,
+                )
+            ) from err
+
+        # Extract all unique ids from vals_list
+        line_ids = {vals.get(key_name) for vals in vals_list if vals.get(key_name)}
+
+        # Fetch all line references in a single query
+        lines = CxModel.browse(line_ids)
+        return {line.id: line.reference for line in lines}
```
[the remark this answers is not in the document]
ivs-cetmix: This suggestion is wrong. Current implementation is done according to the Odoo official documentation:
https://www.odoo.com/documentation/16.0/developer/howtos/translations.html#variables

cetmix_tower_server/models/cx_tower_reference_mixin.py
```diff
@@ -186,10 +186,14 @@ def _get_copied_name(self):
...
         counter = 1
+        # Ensures that the generated copy name is unique by
+        # appending a counter until a unique name is found.
         while self.search_count([("name", "=", copy_name)]) > 0:
             counter += 1
             copy_name = _(
-                "%(name)s (copy %(number)s)", name=original_name, number=str(counter)
+                "%(name)s (copy %(number)s)",
+                name=original_name,
+                number=str(counter),
```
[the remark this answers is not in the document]
ivs-cetmix: This suggestion is wrong. Current implementation is done according to the Odoo official documentation:
https://www.odoo.com/documentation/16.0/developer/howtos/translations.html#variables

cetmix_tower_server/models/cx_tower_reference_mixin.py
ivs-cetmix: Ok, understood

ivs-cetmix (changes requested):
Please add the 'reference` field 'optional=hide' after the name field:
- for plan lines
image

- for plan line actions
image

- for post run action variables
image

Aslo change 'variable_reference' field of the model 'cx.variable.value' model to "Variable Reference" (just remove the 'string=...' in the model)

Aldeigja (approved):
Functional review LGTM

ivs-cetmix (approved):
LGTM

CetmixGitDrone:

CetmixGitDrone:
What a great day to merge this nice PR. Let's do it!
Prepared branch [14.0-dev-ocabot-merge-pr-129-by-ivs-cetmix-bump-patch](https://github.com/cetmix/cetmix-tower/commits/14.0-dev-ocabot-merge-pr-129-by-ivs-cetmix-bump-patch), awaiting test results.

CetmixGitDrone:
Congratulations, your PR was merged at 6a992c06a71dcb5d751432f117d3d503376b6863. Thanks a lot for contributing to cetmix. ❤️

PR #147, opened 2024-11-23
[14.0][FIX] cetmix_tower_server: Link secrets to records

dessanhemrayev:
This commit adds m2m field(secret) to Files and Commands.

Changes

Add field to Files and Commands models
Add field to Files and Commands views

Task: 4042

## Summary by CodeRabbit

- **New Features**
	- Introduced a `secret_ids` field in both `CxTowerCommand` and `CxTowerFile` models, allowing for the association of multiple secrets with commands and files.
	- Enhanced views to include the `secret_ids` field in form and tree displays, with visibility controlled by user permissions.
	- Added a new abstract model `CxTowerKeyMixin` for managing secret relationships and access control.

- **Bug Fixes**
	- Improved error handling and validation for key extraction and resolution processes.

- **Tests**
	- Added new tests to verify the functionality of secret management within commands and files, ensuring proper linking and removal of secret references.

- **Documentation**
	- Updated comments and docstrings for clarity in new methods and classes.

### DIFF
--- a/cetmix_tower_server/models/cx_tower_command.py
+++ b/cetmix_tower_server/models/cx_tower_command.py
@@ -43,6 +43,7 @@ class CxTowerCommand(models.Model):
         "cx.tower.template.mixin",
         "cx.tower.reference.mixin",
         "cx.tower.access.mixin",
+        "cx.tower.key.mixin",
     ]
     _description = "Cetmix Tower Command"
     _order = "name"
--- a/cetmix_tower_server/models/cx_tower_file.py
+++ b/cetmix_tower_server/models/cx_tower_file.py
@@ -30,7 +30,12 @@ INTERVAL_TYPES = {
 
 class CxTowerFile(models.Model):
     _name = "cx.tower.file"
-    _inherit = ["cx.tower.template.mixin", "mail.thread", "mail.activity.mixin"]
+    _inherit = [
+        "cx.tower.template.mixin",
+        "mail.thread",
+        "mail.activity.mixin",
+        "cx.tower.key.mixin",
+    ]
     _description = "Cx Tower File"
     _order = "name"
 
--- a/cetmix_tower_server/models/cx_tower_key.py
+++ b/cetmix_tower_server/models/cx_tower_key.py
@@ -347,20 +347,33 @@ class CxTowerKey(models.Model):
             str: key value or None if not able to parse
         """
 
+        key_parts = self._extract_key_parts(key_string)
+        if key_parts is None:
+            return None
+
+        key_type, reference = key_parts
+        key_value = self._resolve_key(key_type, reference, **kwargs)
+
+        return key_value
+
+    def _extract_key_parts(self, key_string):
+        """Extract and validate key parts from the key string.
+
+        Args:
+            key_string (str): key string
+
+        Returns:
+            tuple: (key_type, reference) if valid, else None
+        """
         key_parts = (
             key_string.replace(" ", "").replace(self.KEY_TERMINATOR, "").split(".")
         )
 
         # Must be 3 parts including pre!
-        if len(key_parts) != 3 or key_parts[0] != self.KEY_PREFIX:
-            return
+        if len(key_parts) == 3 and key_parts[0] == self.KEY_PREFIX:
+            return key_parts[1], key_parts[2]
 
-        key_type = key_parts[1]
-        reference = key_parts[2]
-
-        key_value = self._resolve_key(key_type, reference, **kwargs)
-
-        return key_value
+        return None
 
     def _resolve_key(self, key_type, reference, **kwargs):
         """Resolve key
--- a/cetmix_tower_server/models/cx_tower_key_mixin.py
+++ b/cetmix_tower_server/models/cx_tower_key_mixin.py
@@ -1,0 +1,61 @@
+from odoo import api, fields, models
+
+
+class CxTowerKeyMixin(models.AbstractModel):
+    _name = "cx.tower.key.mixin"
+    _description = "Mixin for managing secrets"
+
+    secret_ids = fields.Many2many(
+        comodel_name="cx.tower.key",
+        compute="_compute_secret_ids",
+        compute_sudo=True,
+        readonly=True,
+        store=True,
+        string="Secrets",
+    )
+
+    @api.depends("code")
+    def _compute_secret_ids(self):
+        """
+        Compute the secret IDs based on the references found in the code field.
+
+        This method updates the secret_ids Many2many field by extracting secret
+        references from the code field. If no code is present, the field is cleared.
+        It ensures updates are only triggered when there are differences between
+        the current and new secret IDs.
+        """
+        for record in self:
+            if record.code:
+                new_secrets = self._extract_secret_ids(record.code)
+
+                # This will create a recordset that contains the difference
+                if record.secret_ids != new_secrets:
+                    record.secret_ids = new_secrets
+            else:
+                record.secret_ids = [(5, 0, 0)]
+
+    @api.model
+    def _extract_secret_ids(self, code):
+        """
+        Extract secret IDs based on references found in the given `code`.
+
+        Args:
+            code: Text containing potential secret references.
+
+        Returns:
+            list: List of secret IDs corresponding to the references in `code`.
+        """
+        key_model = self.env["cx.tower.key"]
+        key_strings = key_model._extract_key_strings(code)
+
+        key_refs = []
+        for key_string in key_strings:
+            key_parts = key_model._extract_key_parts(key_string)
+            if key_parts:
+                key_refs.append(key_parts[1])
+
+        return key_model.search(
+            [
+                ("reference", "in", key_refs),
+            ]
+        )
--- a/cetmix_tower_server/views/cx_tower_command_view.xml
+++ b/cetmix_tower_server/views/cx_tower_command_view.xml
@@ -74,6 +74,11 @@
                                 options="{'color_field': 'color'}"
                             />
                             <field
+                                name="secret_ids"
+                                widget="many2many_tags"
+                                groups="cetmix_tower_server.group_manager,cetmix_tower_server.group_root"
+                            />
+                            <field
                                 name="tag_ids"
                                 widget="many2many_tags"
                                 options="{'color_field': 'color'}"
@@ -172,6 +177,12 @@
                     name="os_ids"
                     widget="many2many_tags"
                     options="{'color_field': 'color'}"
+                    optional="hide"
+                />
+                <field
+                    name="secret_ids"
+                    widget="many2many_tags"
+                    groups="cetmix_tower_server.group_manager,cetmix_tower_server.group_root"
                     optional="hide"
                 />
                 <field name="allow_parallel_run" optional="hide" />
--- a/cetmix_tower_server/views/cx_tower_file_view.xml
+++ b/cetmix_tower_server/views/cx_tower_file_view.xml
@@ -68,6 +68,11 @@
                                 name="keep_when_deleted"
                                 attrs="{'invisible': [('source', '!=', 'tower')]}"
                             />
+                            <field
+                                name="secret_ids"
+                                widget="many2many_tags"
+                                groups="cetmix_tower_server.group_manager,cetmix_tower_server.group_root"
+                            />
                             </group>
                             <group>
                                 <field name="auto_sync" />
@@ -158,7 +163,12 @@
                 <field name="full_server_path" optional="hide" />
                 <field name="sync_date_last" />
                 <field name="template_id" optional="show" />
-
+                <field
+                    name="secret_ids"
+                    widget="many2many_tags"
+                    groups="cetmix_tower_server.group_manager,cetmix_tower_server.group_root"
+                    optional="hide"
+                />
             </tree>
         </field>
     </record>
--- a/cetmix_tower_server/models/__init__.py
+++ b/cetmix_tower_server/models/__init__.py
@@ -4,6 +4,7 @@ from . import cx_tower_variable_mixin
 from . import cx_tower_template_mixin
 from . import cx_tower_access_mixin
 from . import cx_tower_reference_mixin
+from . import cx_tower_key_mixin
 from . import cx_tower_variable
 from . import cx_tower_variable_value
 from . import cx_tower_file
--- a/cetmix_tower_server/tests/test_command.py
+++ b/cetmix_tower_server/tests/test_command.py
@@ -1131,3 +1131,46 @@ else:
             command_result["error"],
             "The error in command result must be None",
         )
+
+    def test_command_with_secret(self):
+        """
+        Test case to verify that when a command includes a secret reference,
+        the secret key is automatically linked with the command.
+        """
+
+        # Command with a secret reference
+        code = "cd {{ test_path_ }} && mkdir #!cxtower.secret.FOLDER!#"
+
+        secrets = self.Command._extract_secret_ids(code)
+        secret_folder_key = self.secret_folder_key
+        self.assertIn(
+            secret_folder_key,
+            secrets,
+            msg=(
+                f"The expected secret ID #{secret_folder_key.id} "
+                "was not found in the provided code."
+            ),
+        )
+
+        command_with_keys = self.Command.create(
+            {"name": "Command with keys", "code": code}
+        )
+
+        # Assert that the secret key is linked with the command
+        self.assertIn(
+            secret_folder_key,
+            command_with_keys.secret_ids,
+            msg="The secret key is not linked with the command.",
+        )
+
+        # Update the command's code to remove the secret reference
+        code = "cd {{ test_path_ }} && mkdir new_folder"
+        command_with_keys.code = code
+
+        self.assertFalse(
+            command_with_keys.secret_ids,
+            msg=(
+                "The secret_ids field should be empty after "
+                "removing the secret reference from command."
+            ),
+        )
--- a/cetmix_tower_server/tests/test_file.py
+++ b/cetmix_tower_server/tests/test_file.py
@@ -336,3 +336,59 @@ class TestTowerFile(TestTowerCommon):
                     "server_dir": "/var/tmp",
                 }
             )
+
+    def test_file_with_secret_key(self):
+        """
+        Test case to verify that when a file includes a secret reference,
+        the secret key is automatically linked with the file.
+        """
+
+        # Create a secret key
+        secret_python_key = self.Key.create(
+            {
+                "name": "python",
+                "reference": "PYTHON",
+                "secret_value": "secretPythonCode",
+                "key_type": "s",
+            }
+        )
+
+        # Create a file template with a reference to the secret key
+        file_template = self.env["cx.tower.file.template"].create(
+            {
+                "name": "Test",
+                "file_name": "test.txt",
+                "server_dir": "/var/tmp",
+                "code": "Please use this secret #!cxtower.secret.PYTHON!#",
+            }
+        )
+
+        # Create a file from the file template
+        file = file_template.create_file(
+            server=self.server_test_1, server_dir="/var/tmp/custom"
+        )
+
+        # Assert that the file's code matches the file template's code
+        self.assertEqual(
+            file.code,
+            file_template.code,
+            msg="The file's code does not match the file template's code.",
+        )
+
+        # Assert that the secret key is associated with the file
+        self.assertIn(
+            secret_python_key,
+            file.secret_ids,
+            msg="The secret key is not associated with the file.",
+        )
+
+        # Update the file's code to remove the secret reference
+        file.code = "Only text"
+
+        self.assertFalse(
+            file.secret_ids,
+            msg=(
+                "The secret_ids field should be empty after "
+                "removing the secret reference from file."
+            ),
+        )


cetmix_tower_server/views/cx_tower_file_view.xml
```diff
@@ -156,7 +157,7 @@
                 <field name="full_server_path" optional="hide" />
                 <field name="sync_date_last" />
                 <field name="template_id" optional="show" />
-
+                <field name="key_ids" widget="many2many_tags" optional="hide" />
```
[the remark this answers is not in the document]
ivs-cetmix: Groups are already defined at the model level. No need to add them in the view again.

cetmix_tower_server/models/cx_tower_command.py
```diff
@@ -86,7 +86,14 @@ def _selection_action(self):
...
     note = fields.Text()
-
+    key_ids = fields.Many2many(
+        comodel_name="cx.tower.key",
+        relation="cx_tower_key_command_rel",
+        column1="command_id",
+        column2="key_id",
+        string="Keys",
+        readonly=True,
+    )
```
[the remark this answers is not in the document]
ivs-cetmix: Many2may fields are not indexed and they don't have the "ondelete" parameter.

cetmix_tower_server/models/cx_tower_command.py
```diff
@@ -147,6 +157,32 @@ def _compute_code(self):
...
+                code_references = set(command.code.split())
+                secret_ids = [
+                    reference_to_secret[ref]
+                    for ref in code_references
+                    if ref in reference_to_secret
+                ]
+                command.secret_ids = [(6, 0, secret_ids)]
+            else:
+                command.secret_ids = False
+
```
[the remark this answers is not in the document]
dessanhemrayev: Ok, but there are some controversial points:
1. In the model with secrets there is no field active
2. The secret always has a reference and there is no need to check it

cetmix_tower_server/models/cx_tower_file.py
```diff
@@ -132,6 +132,43 @@ class CxTowerFile(models.Model):
...
+                # Use a set for faster membership testing
+                code_references = set(command.code.split())
+                secret_ids = [
+                    reference_to_secret[ref]
+                    for ref in code_references
+                    if ref in reference_to_secret
+                ]
+                command.secret_ids = [(6, 0, secret_ids)]
+            else:
+                command.secret_ids = False
```
GabbasovDinar: Duplicating the same method across two models increases maintenance overhead and reduces code clarity. Instead, create a single shared method in a mixin or in "cx.tower.key" model and call it from both models. This ensures consistency and makes future updates easier.

GabbasovDinar (changes requested):

cetmix_tower_server/models/cx_tower_file.py
dessanhemrayev: Thanks for review
I'll fix it

cetmix_tower_server/models/cx_tower_secret_mixin.py
```diff
@@ -0,0 +1,42 @@
...
+                # Use a set for faster membership testing
+                code_references = set(record.code.split())
+                secret_ids = [
+                    reference_to_secret[ref]
+                    for ref in code_references
+                    if ref in reference_to_secret
+                ]
+                record.secret_ids = [(6, 0, secret_ids)]
+            else:
+                record.secret_ids = False
```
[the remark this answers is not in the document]
dessanhemrayev: okay,I'll fix it

GabbasovDinar (approved):

GabbasovDinar:
@dessanhemrayev please, fix conflicts

cetmix_tower_server/models/cx_tower_file.py
```diff
@@ -132,6 +137,11 @@ class CxTowerFile(models.Model):
         compute="_compute_variable_ids",
         store=True,
     )
+    secret_ids = fields.Many2many(
+        comodel_name="cx.tower.key",
+        relation="cx_tower_secret_file_rel",
+        column1="file_id",
+    )
```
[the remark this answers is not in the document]
dessanhemrayev: we are overriding the field from the secret.mixin model and declared these parameters in that model, so there is no need to specify this because when overriding the field only the necessary changes

cetmix_tower_server/models/cx_tower_secret_mixin.py
```diff
@@ -0,0 +1,60 @@
+from odoo import api, fields, models
+
+
+class CxTowerSecretMixin(models.AbstractModel):
+    _name = "cx.tower.secret.mixin"
+    _description = "Mixin for managing secrets"
+
+    secret_ids = fields.Many2many(
+        comodel_name="cx.tower.key",
+        relation="cx_tower_secret_rel",
```
ivs-cetmix: Model A inherits this mixin. New table with the name `cx_tower_secret_rel` is created in the database.
Model B inherits this mixin. New table with the name `cx_tower_secret_rel` is created in the database.
Why?
Do not put explicit relation keys in the mixin.

cetmix_tower_server/models/cx_tower_secret_mixin.py
```diff
@@ -0,0 +1,60 @@
...
+            code: Text containing potential secret references.
+
+        Returns:
+            list: List of secret IDs corresponding to the references in `code`.
+        """
+        if not code:
+            return []
+
+        # Pre-fetch all secrets and map references to IDs
+        secrets = self.env["cx.tower.key"].search([])
```
ivs-cetmix: What if there are 100000 million secrets in the db? Does it mean we need to fetch all of them every time?

cetmix_tower_server/models/cx_tower_secret_mixin.py
```diff
@@ -0,0 +1,60 @@
...
+        reference_to_secret = {
+            f"#!cxtower.secret.{secret.reference}!#": secret.id for secret in secrets
+        }
+
+        # Extract unique references from `code` and map them to secret IDs
+        code_references = set(code.split())
```
ivs-cetmix: What if a secret consist of more or less than 3 parts? 

There is an algorithm and functions defined in the [cx.tower.key](https://github.com/cetmix/cetmix-tower/blob/14.0-dev/cetmix_tower_server/models/cx_tower_key.py) model. You should use them for secret parsing.

cetmix_tower_server/models/cx_tower_secret_mixin.py
```diff
@@ -0,0 +1,60 @@
+from odoo import api, fields, models
+
+
+class CxTowerSecretMixin(models.AbstractModel):
+    _name = "cx.tower.secret.mixin"
```
ivs-cetmix: Secrets and keys are implemented in the cx.tower.key model. So this model name should use the same name pattern.

cetmix_tower_server/views/cx_tower_file_view.xml
```diff
@@ -68,6 +68,7 @@
                                 name="keep_when_deleted"
                                 attrs="{'invisible': [('source', '!=', 'tower')]}"
                             />
+                            <field name="secret_ids" widget="many2many_tags" />
```
ivs-cetmix: Please make this field visible only to the members of the "Manager" and "Root" groups.

cetmix_tower_server/views/cx_tower_file_view.xml
```diff
@@ -158,7 +159,7 @@
                 <field name="full_server_path" optional="hide" />
                 <field name="sync_date_last" />
                 <field name="template_id" optional="show" />
-
+                <field name="secret_ids" widget="many2many_tags" optional="hide" />
```
ivs-cetmix: Please make this field visible only to the members of the "Manager" and "Root" groups.

cetmix_tower_server/views/cx_tower_command_view.xml
```diff
@@ -164,6 +165,7 @@
                     options="{'color_field': 'color'}"
                     optional="hide"
                 />
+                <field name="secret_ids" widget="many2many_tags" optional="hide" />
```
ivs-cetmix: Please make this field visible only to the members of the "Manager" and "Root" groups.

cetmix_tower_server/views/cx_tower_command_view.xml
```diff
@@ -67,6 +67,7 @@
                                 widget="many2many_tags"
                                 options="{'color_field': 'color'}"
                             />
+                            <field name="secret_ids" widget="many2many_tags" />
```
ivs-cetmix: Please make this field visible only to the members of the "Manager" and "Root" groups.

ivs-cetmix (changes requested):

cetmix_tower_server/models/cx_tower_secret_mixin.py
dessanhemrayev: thanks, I'll fix it

cetmix_tower_server/models/cx_tower_secret_mixin.py
dessanhemrayev: thanks, I'll fix it

cetmix_tower_server/models/cx_tower_secret_mixin.py
dessanhemrayev: thanks, I'll fix it

cetmix_tower_server/models/cx_tower_secret_mixin.py
dessanhemrayev: ok, understood

cetmix_tower_server/views/cx_tower_file_view.xml
dessanhemrayev: ok, but groups are already defined at the model level. Isn't that enough?

cetmix_tower_server/models/cx_tower_key_mixin.py
```diff
@@ -0,0 +1,56 @@
...
+        key_model = self.env["cx.tower.key"]
+        key_strings = key_model._extract_key_strings(code)
+
+        key_refs = []
+        for key_string in key_strings:
+            key_parts = key_model._extract_key_parts(key_string)
+            if key_parts:
+                key_refs.append(key_parts[1])
+
+        return key_model.search([("reference", "in", key_refs)]).ids
```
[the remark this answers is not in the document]
dessanhemrayev: 1)I don't think it's a good thing to limit code length.
2) in cx.tower.key model has not  "active" field
3) I thought about adding a new method to check the parts, but I don't think it's the best idea from a performance perspective.

cetmix_tower_server/models/cx_tower_command.py
```diff
@@ -86,7 +87,11 @@ def _selection_action(self):
         string="OSes",
     )
     note = fields.Text()
-
+    secret_ids = fields.Many2many(
+        comodel_name="cx.tower.key",
+        relation="cx_tower_secret_command_rel",
+        column1="command_id",
+    )
```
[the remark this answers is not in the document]
dessanhemrayev: we are overriding the field from the cx.tower.key.mixin model and declared these parameters in that model, so there is no need to specify this because when overriding the field only the necessary changes

cetmix_tower_server/models/cx_tower_key_mixin.py
```diff
@@ -1,13 +1,12 @@
...
-class CxTowerSecretMixin(models.AbstractModel):
-    _name = "cx.tower.secret.mixin"
+class CxTowerKeyMixin(models.AbstractModel):
+    _name = "cx.tower.key.mixin"
     _description = "Mixin for managing secrets"
 
     secret_ids = fields.Many2many(
         comodel_name="cx.tower.key",
-        relation="cx_tower_secret_rel",
         column1="res_id",
```
ivs-cetmix: No relation fields are needed. Just a comodel.

cetmix_tower_server/models/cx_tower_command.py
```diff
@@ -86,7 +87,11 @@ def _selection_action(self):
         string="OSes",
     )
     note = fields.Text()
-
+    secret_ids = fields.Many2many(
```
ivs-cetmix: This field is added by mixin. No need to add it in the model explicitly.

cetmix_tower_server/models/cx_tower_file.py
```diff
@@ -132,6 +137,11 @@ class CxTowerFile(models.Model):
         compute="_compute_variable_ids",
         store=True,
     )
+    secret_ids = fields.Many2many(
```
ivs-cetmix: This field is added by mixin. No need to add it in the model explicitly.

cetmix_tower_server/models/cx_tower_key_mixin.py
```diff
@@ -0,0 +1,56 @@
...
+
+    secret_ids = fields.Many2many(
+        comodel_name="cx.tower.key",
+        column1="res_id",
+        column2="secret_id",
+        string="Secrets",
+        readonly=True,
+        store=True,
+        compute="_compute_secret_ids",
+        groups="cetmix_tower_server.group_manager,cetmix_tower_server.group_root",
```
ivs-cetmix: This field must be be accessible for all users. You need to restrict it's visibility in views.

cetmix_tower_server/models/cx_tower_key_mixin.py
```diff
@@ -0,0 +1,56 @@
...
+        Compute the secret IDs based on the references found in the code field.
+        """
+
+        for record in self:
+            if record.code:
+                # Use a set for faster membership testing
+                secret_ids = self._extract_secret_ids(record.code)
+                record.secret_ids = [(6, 0, secret_ids)]
+            else:
+                record.secret_ids = [(6, 0, [])]
```
ivs-cetmix: Use the [command](https://www.odoo.com/documentation/13.0/uk/developer/reference/addons/orm.html#odoo.models.Model.write) that removes records. Not the one that add them.

cetmix_tower_server/models/cx_tower_key_mixin.py
```diff
@@ -0,0 +1,56 @@
...
+        """
+        Extract secret IDs based on references found in the given `code`.
+
+        Args:
+            code: Text containing potential secret references.
+
+        Returns:
+            list: List of secret IDs corresponding to the references in `code`.
+        """
+        if not code:
```
ivs-cetmix: You can check this before calling this function to avoid unnecessary calls. Also this function resolve model every time which degrades performance. Probably this can be done inside the `compute` function more efficiently.
@GabbasovDinar what do you think?

cetmix_tower_server/views/cx_tower_command_view.xml
ivs-cetmix: Visible != accessible

cetmix_tower_server/views/cx_tower_file_view.xml
ivs-cetmix: We need to restrict visibility, not access.

cetmix_tower_server/views/cx_tower_file_view.xml
ivs-cetmix: Visible != accessible

ivs-cetmix (changes requested):
If you define a field at the mixin level and than re-define it in every model that inherits the mixin, then what's the use of adding that field in the mixing?

cetmix_tower_server/models/cx_tower_key_mixin.py
dessanhemrayev: Thanks, I'll fix it

dessanhemrayev:
> **Actionable comments posted: 3**
> 
> 📜 Review details

I'm sorry, but I have to remove these commits as they are no longer needed and could mislead the reviewer.

cetmix_tower_server/models/cx_tower_key_mixin.py
```diff
@@ -0,0 +1,45 @@
...
+    def _compute_secret_ids(self):
+        """
+        Compute the secret IDs based on the references found in the code field.
+
+        This method updates the secret_ids Many2many field by extracting secret
+        references from the code field. If no code is present, the field is cleared.
+        """
+        for record in self:
+            # Clear secrets by default
+            record.secret_ids = [(5, 0, 0)]
```
ivs-cetmix: This will always trigger an update operation, even if it's not needed.
You should check the difference between current secrets and new ones. Add modify this field only in case this difference exists.

ivs-cetmix (changes requested):
I don't see a reason why should those m2m fields are added in the models. It's actually against the mixin logic.
Why not to add this field to the mixin instead? Ok, you cannot use "column" and "relation" in this case. But this is not an issue so far.

cetmix_tower_server/models/cx_tower_key_mixin.py
dessanhemrayev: Thanks, understood

cetmix_tower_server/models/cx_tower_key_mixin.py
```diff
@@ -0,0 +1,60 @@
...
+                # Get the current secret IDs
+                current_secret_ids = set(record.secret_ids.ids)
+
+                # Update only if there's a difference
+                if new_secret_ids != current_secret_ids:
+                    record.secret_ids = [(4, sid, 0) for sid in new_secret_ids]
+
+            else:
+                record.secret_ids = [(5, 0, 0)]
+
```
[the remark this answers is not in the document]
dessanhemrayev: This will always trigger an update operation, even if it's not needed.

cetmix_tower_server/models/cx_tower_key.py
```diff
@@ -331,20 +331,33 @@ def _parse_key_string(self, key_string, **kwargs):
...
+
+        Returns:
+            tuple: (key_type, reference) if valid, else None
+        """
         key_parts = (
             key_string.replace(" ", "").replace(self.KEY_TERMINATOR, "").split(".")
         )
 
         # Must be 3 parts including pre!
         if len(key_parts) != 3 or key_parts[0] != self.KEY_PREFIX:
```
ivs-cetmix: You should use positive logic here to avoid another explicit return:
```python
if len(key_parts) == 3 and key_parts[0] == self.KEY_PREFIX:
    return key_parts[1], key_parts[2]
```

cetmix_tower_server/models/cx_tower_key.py
```diff
@@ -331,20 +331,33 @@ def _parse_key_string(self, key_string, **kwargs):
...
-            return
+            return None
```
ivs-cetmix: `return None` is the same as `return`. Both will return `None`.

cetmix_tower_server/models/cx_tower_key_mixin.py
```diff
@@ -0,0 +1,59 @@
+from odoo import api, fields, models
+
+
+class CxTowerKeyMixin(models.AbstractModel):
+    _name = "cx.tower.key.mixin"
+    _description = "Mixin for managing secrets"
+
+    secret_ids = fields.Many2many(
```
ivs-cetmix, line 8: Why is this field non-stored? It will implicate a whole bunch of calculations on every read. While "code" filed doesn't have a tendency to be updated frequently.

cetmix_tower_server/models/cx_tower_key_mixin.py
```diff
@@ -0,0 +1,59 @@
...
+
+        This method updates the secret_ids Many2many field by extracting secret
+        references from the code field. If no code is present, the field is cleared.
+        It ensures updates are only triggered when there are differences between
+        the current and new secret IDs.
+        """
+        for record in self:
+            if record.code:
+                # Extract the new set of secret IDs
+                new_secret_ids = set(self._extract_secret_ids(record.code))
```
ivs-cetmix: No need to extract ids here as well as compare sets. You can use a recordset operation for that:

```python
# Return a recordset, not ids
new_secrets = set(self._extract_secret_ids(record.code))  

# This will create a recordset that contains the difference
if record.secret_ids - new_secrets:

    # Need to check, this works in Odoo 16, highly likely would work in 14.0 too    
    record.secret_ids = new_secrets

File: cetmix_tower_server/models/cx_tower_key_mixin.py [hunk 29]
```diff
@@ -0,0 +1,59 @@
...
+        key_model = self.env["cx.tower.key"]
+        key_strings = key_model._extract_key_strings(code)
+
+        key_refs = []
+        for key_string in key_strings:
+            key_parts = key_model._extract_key_parts(key_string)
+            if key_parts:
+                key_refs.append(key_parts[1])
+
+        return key_model.search([("reference", "in", key_refs)]).ids
```
ivs-cetmix: We should need in mind, that `reference` field in the `cx.tower.key` model can [contain](https://github.com/cetmix/cetmix-tower/blob/e80208ca0882530797e1471fe7a6095097dcb5a4/cetmix_tower_server/models/cx_tower_key.py#L49) non unique values.
So we should check only those keys, that don't belong to any partner or server.

ivs-cetmix (changes requested):

cetmix_tower_server/models/cx_tower_key.py
dessanhemrayev: thanks, I'll fix it

cetmix_tower_server/models/cx_tower_key.py
dessanhemrayev: thanks, understood

cetmix_tower_server/models/cx_tower_key_mixin.py
dessanhemrayev: thanks, I'll fix it

cetmix_tower_server/models/cx_tower_key_mixin.py
dessanhemrayev: thanks, I'll check and fix it

cetmix_tower_server/models/cx_tower_key_mixin.py
dessanhemrayev: thanks, I'll check and fix it

cetmix_tower_server/models/cx_tower_key_mixin.py
```diff
@@ -0,0 +1,63 @@
...
+        key_model = self.env["cx.tower.key"]
+        key_strings = key_model._extract_key_strings(code)
+
+        key_refs = []
+        for key_string in key_strings:
+            key_parts = key_model._extract_key_parts(key_string)
+            if key_parts:
+                key_refs.append(key_parts[1])
+
+        return key_model.search(
```
ivs-cetmix, line 57: This is good, however we should still take care of the cases where a secret can be a server or a partner tight.
So probably we need to get all secrets, not just without server and partner.

ivs-cetmix (changes requested):

cetmix_tower_server/models/cx_tower_key_mixin.py
dessanhemrayev: Thanks, do you mean we need to use this [method ](https://github.com/cetmix/cetmix-tower/blob/e80208ca0882530797e1471fe7a6095097dcb5a4/cetmix_tower_server/models/cx_tower_key.py#L363)to check secrets?

cetmix_tower_server/views/cx_tower_file_view.xml
```diff
@@ -66,6 +66,7 @@
                                 name="keep_when_deleted"
                                 attrs="{'invisible': [('source', '!=', 'tower')]}"
                             />
+                            <field name="key_ids" widget="many2many_tags" />
```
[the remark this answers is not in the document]
ivs-cetmix: Groups are already defined at the model level. And we need to show all secrets, not just the ones created by current user.

cetmix_tower_server/models/cx_tower_key_mixin.py
ivs-cetmix: No no, I mean we need to get all keys, not just without server or partner. And of course no need to filter them afterwards.  So just `("reference", "in", key_refs)`

ivs-cetmix (changes requested):

cetmix_tower_server/models/cx_tower_key_mixin.py
dessanhemrayev: okay, I'll fix it

ivs-cetmix (approved):
Code review LGTM

CetmixGitDrone:

Aldeigja (approved):
Functional LGTM

CetmixGitDrone:
What a great day to merge this nice PR. Let's do it!
Prepared branch [14.0-dev-ocabot-merge-pr-147-by-ivs-cetmix-bump-nobump](https://github.com/cetmix/cetmix-tower/commits/14.0-dev-ocabot-merge-pr-147-by-ivs-cetmix-bump-nobump), awaiting test results.

CetmixGitDrone:
Congratulations, your PR was merged at 24f4fc0c627c92a83b0ea3527130937ee9d2beee. Thanks a lot for contributing to cetmix. ❤️

PR #154, opened 2024-12-02
[IMP] cetmix_tower_server: "Required" flag for configuration variables in Server Template

tendil:
## Description

This PR introduces comprehensive updates to the `cetmix_tower_server` module, focusing on the implementation and validation of the "Required" flag for configuration variables in **Server Templates** and its correct handling in the **Server Creation Wizard**. 

The logic ensures that required variables are correctly displayed, validated, and processed during server creation, preventing servers from being created with missing or empty required variables.

### Key Changes

1. **Model Enhancements**
   - Introduced the "Required" field for configuration variables in `cx.tower.variable.value`.
   - Implemented logic to validate required variables during server creation:
     - Ensures required variables cannot be left empty.
     - Displays appropriate error messages when required variables are missing or have empty values.

2. **Server Template Updates**
   - Added the "Required" flag to the **Configuration** tab in **Server Templates**.
   - When creating a new server, variables marked as "Required" in the template are correctly passed to the **Server Creation Wizard**.

3. **Server Creation Wizard Enhancements**
   - Updated the wizard to:
     - Display required variables with non-editable checkboxes indicating their "Required" status.
     - Prevent server creation if any required variable is missing or has an empty value.
   - Implemented backend validation:
     - Missing variables trigger the error: `Missing variables: %(variables)s`.
     - Empty values for required variables trigger the error: `Empty values for variables: %(variables)s`.

4. **Error Messages**
   - Improved error handling to provide user-friendly and translatable error messages for missing or empty required variables.

5. **Test Suite**
   - Added automated tests to validate functionality and edge cases:
     - **test_required_flag_in_create_server_wizard**: Verifies that the "Required" flag is passed correctly to the server creation wizard.
     - **test_required_attribute_in_wizard_field**: Ensures required fields are visually marked in the wizard.
     - **test_successful_server_creation_with_required_variables**: Confirms successful server creation when required variables are provided.
     - **test_optional_variable_with_empty_value**: Validates that optional variables can have empty values without blocking server creation.
     - **test_wizard_without_variables**: Tests the wizard's behavior when no variables are configured.
     - **test_update_required_variable_value**: Ensures required variable values can be updated correctly in the wizard.
     - **test_optional_variable_handling**: Confirms proper handling of optional variables across templates and servers.

---

### Steps to Test

1. Navigate to **Servers -> Server Templates** in the **Cetmix Tower** module.
2. Create a new server template:
   - Fill in all mandatory fields on the **General Settings** tab.
   - Go to the **Configuration** tab and add variables.
   - Mark some variables as "Required" by activating the checkbox.
   - Save the template.
3. Click the **Create Server** button to open the **Server Creation Wizard**.
4. Verify the following in the wizard:
   - Required variables are displayed with non-editable "Required" checkboxes.
   - If any required variable is missing a value, the **Create** button is not visible.
   - If a required variable is removed from the list, attempting to create the server triggers the error:
     - `Missing variables: %(variables)s`.
   - If a required variable is present but has an empty value, attempting to create the server triggers the error:
     - `Empty values for variables: %(variables)s`.
5. Fill in all required variables and complete the server creation process.
6. Navigate to **Servers -> Servers** and verify:
   - The new server is created with all required variables saved correctly.
   - Optional variables with empty values are handled properly.

---

### Related Issue

Closes #4070: cetmix_tower_server Server Template: "required" option for configuration variable

---

## Summary by CodeRabbit

- **New Features**
	- Added a `required` field to indicate mandatory configuration variables in server templates and wizards.
	- Enhanced server creation wizards to validate required variables and provide user-friendly messages for missing values.

- **Bug Fixes**
	- Implemented validation to prevent server creation if required variables are missing.

- **Tests**
	- Added comprehensive tests for required and optional variable handling in server creation wizards.

- **Documentation**
	- Updated XML views to include new fields and validation messages for better user experience.

### DIFF
--- a/cetmix_tower_server/models/cx_tower_server_template.py
+++ b/cetmix_tower_server/models/cx_tower_server_template.py
@@ -2,6 +2,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 from odoo import _, api, fields, models
+from odoo.exceptions import ValidationError
 
 
 class CxTowerServerTemplate(models.Model):
@@ -113,6 +114,7 @@ class CxTowerServerTemplate(models.Model):
                             {
                                 "variable_id": line.variable_id.id,
                                 "value_char": line.value_char,
+                                "required": line.required,
                             },
                         )
                         for line in self.variable_value_ids
@@ -293,6 +295,9 @@ class CxTowerServerTemplate(models.Model):
             # custom specific variable values
             configuration_variables = kwargs.pop("configuration_variables", None)
             if configuration_variables:
+                # Validate required variables
+                self._validate_required_variables(configuration_variables)
+
                 variable_vals_list = []
                 variable_obj = self.env["cx.tower.variable"]
 
@@ -348,6 +353,50 @@ class CxTowerServerTemplate(models.Model):
             values.update(kwargs)
 
         return vals_list
+
+    def _validate_required_variables(self, configuration_variables):
+        """
+        Validate that all required variables are present and not empty.
+
+        Args:
+            configuration_variables (dict): A dictionary of variable references
+                                             and their values.
+
+        Raises:
+            ValidationError: If any required variables are missing or empty.
+        """
+        missing_variables = []
+        empty_variables = []
+
+        for variable in self.variable_value_ids.filtered("required"):
+            variable_ref = variable.variable_reference
+            if variable_ref not in configuration_variables:
+                missing_variables.append(variable_ref)
+            elif not configuration_variables[variable_ref]:
+                empty_variables.append(variable_ref)
+
+        # Construct error message if issues are found
+        if missing_variables or empty_variables:
+            self.variable_value_ids = self.variable_value_ids
+            error_message = _(
+                "Please resolve the following issues with configuration variables:\n"
+            )
+
+            if missing_variables:
+                error_message += _(
+                    "  - Missing variables: %(variables)s\n",
+                    variables=", ".join(missing_variables),
+                )
+
+            if empty_variables:
+                error_message += _(
+                    "  - Empty values for variables: %(variables)s",
+                    variables=", ".join(empty_variables),
+                )
+
+            self.env.context = dict(self.env.context, keep_variable_state=True)
+
+            raise ValidationError(error_message)
 
     def copy(self, default=None):
         """Duplicate the server template along with variable values and server logs."""
--- a/cetmix_tower_server/models/cx_tower_variable_value.py
+++ b/cetmix_tower_server/models/cx_tower_variable_value.py
@@ -57,6 +57,8 @@ class TowerVariableValue(models.Model):
         compute="_compute_variable_ids",
         store=True,
     )
+    required = fields.Boolean()
+
     _sql_constraints = [
         (
             "tower_variable_value_uniq",
--- a/cetmix_tower_server/tests/test_server_template.py
+++ b/cetmix_tower_server/tests/test_server_template.py
@@ -295,3 +295,289 @@ class TestTowerServerTemplate(TestTowerCommon):
             original_log.command_id.code,
             "Command code should be the same in the copied server log",
         )
+
+    def test_required_flag_in_create_server_wizard(self):
+        """
+        Test that the 'Required' flag is correctly passed from the Server Template
+        to the New Server Wizard.
+        """
+        # Set the required flag for a variable
+        self.VariableValue.create(
+            {
+                "variable_id": self.variable_version.id,
+                "server_template_id": self.server_template_sample.id,
+                "value_char": "Test Value",
+                "required": True,
+            }
+        )
+
+        # Open the server creation wizard
+        action = self.server_template_sample.action_create_server()
+        wizard_context = action.get("context", {})
+        default_line_ids = wizard_context.get("default_line_ids", [])
+
+        # We check that required is passed to the wizard
+        self.assertTrue(
+            any(line[2].get("required") for line in default_line_ids),
+            "The 'Required' flag should be correctly passed to the wizard.",
+        )
+
+    def test_required_attribute_in_wizard_field(self):
+        """
+        Test that the 'required' attribute
+        is correctly applied to the 'value_char' field
+        in the wizard when the variable is marked as required.
+        """
+        # Create a required variable
+        self.VariableValue.create(
+            {
+                "variable_id": self.variable_version.id,
+                "server_template_id": self.server_template_sample.id,
+                "value_char": "Test Value",
+                "required": True,
+            }
+        )
+
+        # Open the wizard
+        wizard = self.env["cx.tower.server.template.create.wizard"].create(
+            {
+                "server_template_id": self.server_template_sample.id,
+                "name": "Test Server",
+                "ssh_username": "admin",
+            }
+        )
+
+        # Checking that the 'required' flag is passed to the form context
+        required_fields = [
+            line.required
+            for line in wizard.line_ids
+            if line.variable_id == self.variable_version
+        ]
+        self.assertTrue(
+            all(required_fields),
+            "The 'required' attribute should be correctly "
+            "applied to the 'value_char' field for required variables.",
+        )
+
+    def test_successful_server_creation_with_required_variables(self):
+        """
+        Test that a server is successfully created
+        when all required variables are filled in the wizard.
+        """
+        # Adding a required variable
+        self.VariableValue.create(
+            {
+                "variable_id": self.variable_version.id,
+                "server_template_id": self.server_template_sample.id,
+                "value_char": "",
+                "required": True,
+            }
+        )
+
+        # Open the wizard and fill in the data
+        wizard = self.env["cx.tower.server.template.create.wizard"].create(
+            {
+                "server_template_id": self.server_template_sample.id,
+                "name": "Test Server With Required Variables",
+                "ssh_username": "admin",
+                "line_ids": [
+                    (
+                        0,
+                        0,
+                        {
+                            "variable_id": self.variable_version.id,
+                            "value_char": "Test Value",
+                            "required": True,
+                        },
+                    )
+                ],
+            }
+        )
+
+        # Checking the successful creation of the server
+        action = wizard.action_confirm()
+        self.assertTrue(action, "Server should be created successfully.")
+
+        # Checking that the server has been created
+        server = self.Server.search(
+            [
+                ("name", "=", "Test Server With Required Variables"),
+                ("server_template_id", "=", self.server_template_sample.id),
+            ]
+        )
+        self.assertTrue(server, "Server should exist.")
+        self.assertEqual(
+            server.variable_value_ids.filtered(
+                lambda v: v.variable_id == self.variable_version
+            ).value_char,
+            "Test Value",
+            "The variable value should be saved correctly.",
+        )
+
+    def test_optional_variable_with_empty_value(self):
+        """
+        Test that an optional variable
+        with an empty value is saved correctly
+        in the wizard and does not block server creation.
+        """
+        # Adding an optional variable
+        self.VariableValue.create(
+            {
+                "variable_id": self.variable_url.id,
+                "server_template_id": self.server_template_sample.id,
+                "value_char": "",
+                "required": False,
+            }
+        )
+
+        # Open the wizard
+        wizard = self.env["cx.tower.server.template.create.wizard"].create(
+            {
+                "server_template_id": self.server_template_sample.id,
+                "name": "Server With Optional Variable",
+                "ssh_username": "admin",
+                "line_ids": [
+                    (
+                        0,
+                        0,
+                        {
+                            "variable_id": self.variable_url.id,
+                            "value_char": "",
+                            "required": False,
+                        },
+                    )
+                ],
+            }
+        )
+
+        # Checking that the wizard is saved without errors
+        wizard.action_confirm()
+
+        # Checking that the server has been created
+        server = self.Server.search(
+            [
+                ("name", "=", "Server With Optional Variable"),
+                ("server_template_id", "=", self.server_template_sample.id),
+            ]
+        )
+        self.assertTrue(
+            server, "Server should be created successfully with optional variables."
+        )
+
+        # Checking that an optional variable is saved with an empty value
+        variable = server.variable_value_ids.filtered(
+            lambda v: v.variable_id == self.variable_url
+        )
+        self.assertTrue(variable, "Optional variable should be attached to the server.")
+        self.assertEqual(
+            variable.value_char, "", "Optional variable should have an empty value."
+        )
+
+    def test_wizard_without_variables(self):
+        """
+        Test that the wizard does not display
+        any variables if the server template has none.
+        """
+        # Removing all variables from the template
+        self.VariableValue.search(
+            [("server_template_id", "=", self.server_template_sample.id)]
+        ).unlink()
+
+        # Open the wizard
+        wizard = self.env["cx.tower.server.template.create.wizard"].create(
+            {
+                "server_template_id": self.server_template_sample.id,
+                "name": "Server Without Variables",
+                "ssh_username": "admin",
+            }
+        )
+
+        # Checking that the wizard does not contain variables
+        self.assertFalse(wizard.line_ids, "Wizard should not display any variables.")
+
+    def test_update_required_variable_value(self):
+        """
+        Test that the value of a required variable
+        can be updated in the wizard and saved correctly.
+        """
+        # Adding a required variable
+        self.VariableValue.create(
+            {
+                "variable_id": self.variable_version.id,
+                "server_template_id": self.server_template_sample.id,
+                "value_char": "Old Value",
+                "required": True,
+            }
+        )
+
+        # Open the wizard and update the variable value
+        wizard = self.env["cx.tower.server.template.create.wizard"].create(
+            {
+                "server_template_id": self.server_template_sample.id,
+                "name": "Server With Updated Variable",
+                "ssh_username": "admin",
+                "line_ids": [
+                    (
+                        0,
+                        0,
+                        {
+                            "variable_id": self.variable_version.id,
+                            "value_char": "New Value",
+                            "required": True,
+                        },
+                    )
+                ],
+            }
+        )
+        wizard.action_confirm()
+
+        # Checking that the variable value has been updated
+        server = self.Server.search([("name", "=", "Server With Updated Variable")])
+        variable = server.variable_value_ids.filtered(
+            lambda v: v.variable_id == self.variable_version
+        )
+        self.assertEqual(
+            variable.value_char,
+            "New Value",
+            "The variable value should be updated correctly.",
+        )
+
+    def test_optional_variable_handling(self):
+        """
+        Test that optional variables do not block server creation,
+        even if their values are empty or missing.
+        """
+        # Adding an optional variable to the template
+        self.VariableValue.create(
+            {
+                "variable_id": self.variable_url.id,
+                "server_template_id": self.server_template_sample.id,
+                "value_char": "",
+                "required": False,
+            }
+        )
+
+        # Specify an optional variable with an empty value
+        values = self.server_template_sample._prepare_server_values(
+            configuration_variables={self.variable_url.reference: ""}
+        )
+
+        # Checking that the optional variable is processed correctly
+        variable_data = next(
+            (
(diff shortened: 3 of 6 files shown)


cetmix_tower_server/wizards/cx_tower_server_template_create_wizard.py
```diff
@@ -68,8 +69,11 @@ def _prepare_server_parameters(self):
             res.update(
                 {
                     "configuration_variables": {
-                        line.variable_reference: line.value_char
-                        for line in self.line_ids
+                    line.variable_reference: {
+                        "value_char": line.value_char,
+                        "required": line.required,
+                    }
+                    for line in self.line_ids
```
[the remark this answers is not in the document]
tendil: I believe that adding validation to the `action_confirm` method is not an optimal solution in this case. The validation for required fields is already implemented at a higher level — in the `_create_new_server` method. This ensures centralized validation logic and minimizes code duplication.

Adding this logic to `action_confirm` would result in:
1. **Duplication of validation**: The validation logic would exist in two places, making it harder to maintain.
2. **Violation of single responsibility**: The `action_confirm` method is meant to prepare data and invoke server logic, not to perform checks that should already be handled by the server.
3. **Need for additional synchronization**: Any changes to the required field logic would need to be applied in multiple places, increasing the risk of errors and the complexity of maintenance.

Given these points, I suggest keeping the current validation at the `_create_new_server` level, where it is logically appropriate and already performs its task effectively. This approach maintains cleaner code and avoids unnecessary duplication.

cetmix_tower_server/models/cx_tower_server_template.py
```diff
@@ -200,6 +201,36 @@ def _create_new_server(self, name, **kwargs):
...
+                    "required": details,
+                }
+            else:  # For other value types we treat it as an empty string
+                normalized_variables[variable] = {
+                    "value_char": "",
+                    "required": False,
+                }
+
+        kwargs["configuration_variables"] = normalized_variables
+
```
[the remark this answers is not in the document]
tendil: I believe adding this logic directly to this method may introduce a few issues:
1. **Duplication of Validation**:
   - Validation for required variables is already handled in other parts of the system, including at the model level or in methods like `_create_new_server`.
   - Adding validation here increases the risk of duplicating logic, making maintenance and consistency harder.

2. **Misplaced Responsibility**:
   - This method is intended for data normalization, not validation. Validation should be performed at the model level or in invoked methods like `_create_new_server`.

3. **Increased Testing Complexity**:
   - Adding validation logic here introduces additional points that need to be tested, duplicating scenarios already covered at the model level.

#### Conclusion:
The suggestion is valuable, but to maintain architectural cleanliness and ease of maintenance, it is better to leave validation at the business logic level in the invoked methods.

cetmix_tower_server/models/cx_tower_variable_value.py
```diff
@@ -87,6 +89,15 @@ class TowerVariableValue(models.Model):
         ),
     ]
 
+    @api.constrains("required", "value_char")
+    def _check_required_values(self):
+        for record in self.filtered(lambda r: r.required and r.active):
```
ivs-cetmix: "filtered" will perform and extra iteration over the same recorded.
So you can do the same check within the "for ..." cycle to save time and improve performance.

cetmix_tower_server/models/cx_tower_variable_value.py
```diff
@@ -87,6 +89,15 @@ class TowerVariableValue(models.Model):
         ),
     ]
 
+    @api.constrains("required", "value_char")
+    def _check_required_values(self):
```
ivs-cetmix: According to the task required  variable values are used only in the wizard that creates new server from template.
Why do we need this constraint here?

ivs-cetmix (changes requested):

cetmix_tower_server/models/cx_tower_variable_value.py
tendil: You’re right, this constraint is redundant since required variable validation is only used in the server creation wizard. I’ll remove it to avoid unnecessary checks at the model level.

cetmix_tower_server/models/cx_tower_variable_value.py
tendil: You are correct that using filtered adds an extra iteration, which is suboptimal. The filtering logic can be incorporated directly into the for loop to avoid unnecessary operations. I will make the necessary changes to improve performance.

tendil: I’ll keep in mind for the future that using filtered can be inefficient and it’s better to integrate filtering directly into the loop. However, this method is being removed as part of the current changes, so optimization won’t be necessary here.

ivs-cetmix: It's not inefficient, actually this is one of the most efficient methods of the ORM.
However in this case it will iterate the same recordset that you will iterate later. So there is simply no need for it because you can check the same conditions within your loop.

cetmix_tower_server/models/cx_tower_server_template.py
```diff
@@ -200,6 +201,36 @@ def _create_new_server(self, name, **kwargs):
         Returns:
             cx.tower.server: newly created server record
         """
+        configuration_variables = kwargs.get("configuration_variables", {})
+        normalized_variables = {}
+        for variable, details in configuration_variables.items():
```
ivs-cetmix: What are you trying to do here and why?
Please add comments to your code otherwise it's barely understandable.

ivs-cetmix (changes requested):

cetmix_tower_server/views/cx_tower_server_view.xml
```diff
@@ -239,6 +239,7 @@
                                     <field name="variable_reference" invisible="1" />
                                     <field name="variable_id" />
                                     <field name="value_char" />
+                                    <field name="required" readonly="1" />
```
ivs-cetmix: Why is this field added here? There is nothing about it in the task.

cetmix_tower_server/views/cx_tower_server_view.xml
```diff
@@ -252,6 +253,7 @@
                                             />
                                         </div>
                                         <field name="value_char" />
+                                        <field name="required" readonly="1" />
```
ivs-cetmix: Why is this field added here? There is nothing about it in the task.

ivs-cetmix (changes requested):

cetmix_tower_server/views/cx_tower_server_template_view.xml
```diff
@@ -176,12 +176,20 @@
                                 <tree editable="bottom">
                                     <field name="variable_reference" invisible="1" />
                                     <field name="variable_id" />
-                                    <field name="value_char" />
+                                    <field
+                                        name="value_char"
```
ivs-cetmix: According to the task: `For all checked lines this configuration variable must be a mandatory field in the New Server Wizard:`
This is not a Server Wizard.

cetmix_tower_server/views/cx_tower_server_template_view.xml
```diff
@@ -176,12 +176,20 @@
...
+                                    />
+                                    <field name="required" />
                                 </tree>
                                 <form>
                                     <group>
                                         <field name="variable_id" />
-                                        <field name="value_char" />
+                                        <field
+                                            name="value_char"
+                                            attrs="{'required': [('required', '=', True)]}"
```
ivs-cetmix: According to the task: `For all checked lines this configuration variable must be a mandatory field in the New Server Wizard:`
This is not a Server Wizard.

ivs-cetmix (changes requested):

cetmix_tower_server/models/cx_tower_variable_value.py
```diff
@@ -57,6 +57,8 @@ class TowerVariableValue(models.Model):
         compute="_compute_variable_ids",
         store=True,
     )
+    required = fields.Boolean(default=False)
```
ivs-cetmix: No need to set `False` explicitly because this is default value. Don't let AI fool you!))

cetmix_tower_server/wizards/cx_tower_server_template_create_wizard.py
```diff
@@ -47,7 +47,9 @@ class CxTowerServerTemplateCreateWizard(models.TransientModel):
     line_ids = fields.One2many(
         comodel_name="cx.tower.server.template.create.wizard.line",
         inverse_name="wizard_id",
+        string="Configuration Variables",
     )
+    required = fields.Boolean(default=False)
```
ivs-cetmix: Same here

cetmix_tower_server/wizards/cx_tower_server_template_create_wizard.py
```diff
@@ -102,3 +104,4 @@ class CxTowerServerTemplateCreateWizardVariableLine(models.TransientModel):
     variable_id = fields.Many2one(comodel_name="cx.tower.variable", required=True)
     variable_reference = fields.Char(related="variable_id.reference", readonly=True)
     value_char = fields.Char(string="Value")
+    required = fields.Boolean(string="Required", default=False)
```
ivs-cetmix: And here

cetmix_tower_server/tests/test_server_template.py
```diff
@@ -295,3 +295,348 @@ def test_server_template_copy(self):
...
+            {
+                "variable_id": self.variable_version.id,
+                "server_template_id": self.server_template_sample.id,
+                "value_char": "Test Value",
+                "required": True,
+            }
+        )
+
+        # Checking that the variable is saved with the required flag
+        variable = self.VariableValue.search(
```
ivs-cetmix: This is an ORM function which we are not modifying. So no need to test its behaviour because this is done by the ORM tests.

cetmix_tower_server/tests/test_server_template.py
```diff
@@ -295,3 +295,348 @@ def test_server_template_copy(self):
...
+            }
+        )
+
+        # Open the wizard
+        action = self.server_template_sample.action_create_server()
+        wizard_context = action.get("context", {})
+        default_line_ids = wizard_context.get("default_line_ids", [])
+
+        # Checking that the variable is displayed in the context
+        self.assertTrue(
```
ivs-cetmix: This is an ORM function. No need to test it.

cetmix_tower_server/tests/test_server_template.py
```diff
@@ -295,3 +295,348 @@ def test_server_template_copy(self):
...
+                ],
+            }
+        )
+
+        # Checking the successful creation of the server
+        action = wizard.action_confirm()
+        self.assertTrue(action, "Server should be created successfully.")
+
+        # Checking that the server has been created
+        server = self.Server.search([("name", "=", "Test Server")])
```
ivs-cetmix: What if there is another Server with the same name already exists? Or this is a translatable field.
 You should always use explicitly unique fields when searching for a record.

cetmix_tower_server/tests/test_server_template.py
```diff
@@ -295,3 +295,348 @@ def test_server_template_copy(self):
...
+                    )
+                ],
+            }
+        )
+
+        # Checking that the wizard is saved without errors
+        wizard.action_confirm()
+
+        # Checking that the server has been created
+        server = self.Server.search([("name", "=", "Server With Optional Variable")])
```
ivs-cetmix: Same here

cetmix_tower_server/tests/test_server_template.py
```diff
@@ -295,3 +295,348 @@ def test_server_template_copy(self):
...
+            {
+                "variable_id": self.variable_version.id,
+                "server_template_id": self.server_template_sample.id,
+                "value_char": "Duplicate Value",
+                "required": True,
+            }
+        )
+
+        # Duplicate the template
+        copied_template = self.server_template_sample.copy()
```
ivs-cetmix: You are testing an ORM feature that we are not modifying. No need to do that.

ivs-cetmix (changes requested):

cetmix_tower_server/models/cx_tower_variable_value.py
```diff
@@ -57,6 +57,8 @@ class TowerVariableValue(models.Model):
         compute="_compute_variable_ids",
         store=True,
     )
+    required = fields.Boolean()
+
```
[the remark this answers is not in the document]
ivs-cetmix: "Required" means that this value should be used when creating a new server. At the same time this value can remain empty in the server template. Because it will be populated later in the server creation wizard.

Aldeigja (approved):
Functional LGTM

cetmix_tower_server/models/cx_tower_server_template.py
```diff
@@ -260,37 +261,37 @@ def _prepare_server_values(self, **kwargs):
...
+        # Define the magic fields that should not be copied
         # (including ID and concurrency fields)
         MAGIC_FIELDS = models.MAGIC_COLUMNS + [self.CONCURRENCY_CHECK_FIELD]
 
-        # read all values required to create a new server from the template
+        # Read all values required to create a new server from the template
         vals_list = self.read(self._get_fields_tower_server(), load=False)
 
-        # process each template record
+        # Process each template record
```
GabbasovDinar: please don't rewrite existing code (in this case, comments) that doesn't improve or change anything (and it's acceptable to use comments with a small letter) to preserve git's change history (here and elsewhere )

GabbasovDinar (approved):

cetmix_tower_server/models/cx_tower_server_template.py
tendil: Resolve

cetmix_tower_server/tests/test_server_template.py
```diff
@@ -295,3 +295,249 @@ def test_server_template_copy(self):
...
+        action = wizard.action_confirm()
+        self.assertTrue(action, "Server should be created successfully.")
+
+        # Checking that the server has been created
+        server = self.Server.search(
+            [
+                ("name", "=", "Test Server"),
+                ("server_template_id", "=", self.server_template_sample.id),
+            ]
+        )
```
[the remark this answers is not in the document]
tendil: I fully agree that using unique fields, such as reference, for searching makes the tests more robust and resilient to changes in the data (e.g. possible duplicate names). I have already implemented this recommendation and updated the relevant lines of code.

cetmix_tower_server/models/cx_tower_server_template.py
```diff
@@ -349,6 +354,46 @@ def _prepare_server_values(self, **kwargs):
...
+
+        # Construct error message if issues are found
+        if missing_variables or empty_variables:
+            self.variable_value_ids = self.variable_value_ids
+            error_message = _(
+                "Please resolve the following issues with configuration variables:\n"
+            )
+            if missing_variables:
+                error_message += _(
+                    f"  - Missing variables: {', '.join(missing_variables)}\n"
```
ivs-cetmix: This construction may create issues with translations. Please use the recommended Odoo string translation approach. Check the DevMemo for details.

cetmix_tower_server/models/cx_tower_server_template.py
```diff
@@ -349,6 +354,46 @@ def _prepare_server_values(self, **kwargs):
...
+                )
+            if empty_variables:
+                error_message += _(
+                    f"  - Empty values for variables: {', '.join(empty_variables)}"
```
ivs-cetmix: Same here.

cetmix_tower_server/tests/test_server_template.py
```diff
@@ -295,3 +295,289 @@ def test_server_template_copy(self):
...
+
+        # Checking the successful creation of the server
+        action = wizard.action_confirm()
+        self.assertTrue(action, "Server should be created successfully.")
+
+        # Checking that the server has been created
+        server = self.Server.search(
+            [
+                ("reference", "=", "test_server"),
```
ivs-cetmix: This reference in too generic and can be shadowed with some other Server. So we need either to pass reference explicitly or use more distinguishable name.

cetmix_tower_server/wizards/cx_tower_server_template_create_wizard.py
```diff
@@ -47,7 +48,53 @@ class CxTowerServerTemplateCreateWizard(models.TransientModel):
...
+        store=False,
+    )
+    missing_required_variables = fields.Text(
+        string="Missing Required Variables",
+        compute="_compute_missing_required_variables",
+    )
+    missing_required_variables_message = fields.Text(
+        string="Missing Variables Message",
+        compute="_compute_missing_required_variables_message",
+        store=False,
```
ivs-cetmix: No need to set it to False explicitly, because it's the default value.

cetmix_tower_server/wizards/cx_tower_server_template_create_wizard.py
```diff
@@ -47,7 +48,53 @@ class CxTowerServerTemplateCreateWizard(models.TransientModel):
...
         inverse_name="wizard_id",
+        string="Configuration Variables",
     )
+    required = fields.Boolean(
+        string="Required", help="Indicates if this configuration is mandatory"
+    )
+    has_missing_required_values = fields.Boolean(
+        string="Has Missing Required Values",
+        compute="_compute_has_missing_required_values",
+        store=False,
```
ivs-cetmix: No need to set it to False explicitly, because it's the default value.

cetmix_tower_server/wizards/cx_tower_server_template_create_wizard.py
```diff
@@ -47,7 +48,53 @@ class CxTowerServerTemplateCreateWizard(models.TransientModel):
     line_ids = fields.One2many(
         comodel_name="cx.tower.server.template.create.wizard.line",
         inverse_name="wizard_id",
+        string="Configuration Variables",
     )
+    required = fields.Boolean(
+        string="Required", help="Indicates if this configuration is mandatory"
```
ivs-cetmix: Indicates if this configuration parameter is mandatory

cetmix_tower_server/wizards/cx_tower_server_template_create_wizard.py
```diff
@@ -47,7 +48,53 @@ class CxTowerServerTemplateCreateWizard(models.TransientModel):
...
+
+    @api.depends("has_missing_required_values")
+    def _compute_missing_required_variables_message(self):
+        """
+        Computes the user-friendly message for missing required variables.
+        """
+        for wizard in self:
+            if wizard.has_missing_required_values and wizard.missing_required_variables:
+                wizard.missing_required_variables_message = (
+                    f"Please provide values for the following configuration variables: "
```
ivs-cetmix: You should use translatable strings here. Please check the DevMemo for details.

ivs-cetmix (changes requested):

cetmix_tower_server/wizards/cx_tower_server_template_create_wizard.py
```diff
@@ -47,7 +48,52 @@ class CxTowerServerTemplateCreateWizard(models.TransientModel):
     line_ids = fields.One2many(
         comodel_name="cx.tower.server.template.create.wizard.line",
         inverse_name="wizard_id",
+        string="Configuration Variables",
     )
+    required = fields.Boolean(
+        string="Required", help="Indicates if this configuration parameter is mandatory"
+    )
```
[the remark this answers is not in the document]
tendil: The required field at the wizard level was indeed redundant and misplaced. Since the requirement flag is a property of individual configuration variables (handled at the line model level), it makes sense to remove it from the wizard class.

The field removal has no impact on functionality as the requirement validation already occurs on the line_ids level.

Resolution:
The redundant required field has been removed from the wizard class. The line model correctly handles the "required" logic as intended.

ivs-cetmix (approved):
Code review LGTM

CetmixGitDrone:

CetmixGitDrone:
This PR looks fantastic, let's merge it!
Prepared branch [14.0-dev-ocabot-merge-pr-154-by-ivs-cetmix-bump-nobump](https://github.com/cetmix/cetmix-tower/commits/14.0-dev-ocabot-merge-pr-154-by-ivs-cetmix-bump-nobump), awaiting test results.

CetmixGitDrone:
Congratulations, your PR was merged at 67476e08fd4225d3357ad3c06ffcad230d23181f. Thanks a lot for contributing to cetmix. ❤️

PR #158, opened 2024-12-11
[14.0][IMP] cetmix_tower_server: Improve secrets/keys handling

ivs-cetmix:
Following features are implemented, each on them in a separate commit:

- Improve the way related keys are stored in commands. 
Store only keys that either global or connected to command servers or to partners connected to command servers.
- Implement related secret storage for the file templates.
- Modify cx.tower.key constraints to ensure that they are working as expected.
SQL constraint was useless because both 'partner' and 'server' fields allow NULL values.
So an ORM constraint was added to ensure that the key is unique.

Note for code reviewers: it may be more convenient to do code review for each commit separately.

## Release Notes

- **New Features**
	- Added a `server_status` field to enhance command execution tracking.
	- Introduced a `secret_ids` field in the file template form view for managing secret associations.

- **Bug Fixes**
	- Improved validation for unique key references to prevent duplicates.

- **Documentation**
	- Updated URLs in the README and HTML documentation to reference the stable branch.
	- Standardized formatting for better readability in documentation.

- **Tests**
	- Enhanced test cases to validate the handling of secret keys and unique constraints.
	- Improved clarity in existing test cases.

- **Chores**
	- Minor updates to error handling in YAML processing methods.
	- Added comment directives for static code analysis.

### DIFF
--- a/cetmix_tower_server/__init__.py
+++ b/cetmix_tower_server/__init__.py
@@ -1,2 +1,4 @@
+# pylint: disable=E8103
+
 from . import models
 from . import wizards
--- a/cetmix_tower_server/models/cx_tower_key.py
+++ b/cetmix_tower_server/models/cx_tower_key.py
@@ -3,6 +3,7 @@
 import re
 
 from odoo import _, api, fields, models
+from odoo.exceptions import ValidationError
 from odoo.osv.expression import OR
 
 
@@ -47,13 +48,49 @@ class CxTowerKey(models.Model):
     )
     note = fields.Text()
 
-    _sql_constraints = [
-        (
-            "reference_unique",
-            "UNIQUE(reference, partner_id, server_id)",
-            "Reference must be unique",
+    def init(self):
+        """
+        Remove constraint defined from the mixin
+        """
+        self._cr.execute(
+            """
+            DO $$
+            BEGIN
+                IF EXISTS (
+                    SELECT 1
+                    FROM information_schema.table_constraints
+                    WHERE table_name = 'cx_tower_key' AND
+                    constraint_name = 'cx_tower_key_reference_unique'
+                ) THEN
+                    ALTER TABLE cx_tower_key
+                    DROP CONSTRAINT cx_tower_key_reference_unique;
+                END IF;
+            END$$;
+            """
         )
-    ]
+
+    @api.constrains("reference", "partner_id", "server_id")
+    def _check_reference_unique(self):
+        """ORM constraint to ensure uniqueness"""
+
+        # Need to check archive records as well
+        for rec in self.with_context(active_test=False):
+            if (
+                self.search_count(
+                    [
+                        ("reference", "=", rec.reference),
+                        ("partner_id", "=", rec.partner_id.id),
+                        ("server_id", "=", rec.server_id.id),
+                    ]
+                )
+                > 1
+            ):
+                raise ValidationError(
+                    _(
+                        "Reference must be unique for the combination of partner"
+                        " and server"
+                    )
+                )
 
     def _compute_reference_code(self):
         """Compute key reference
--- a/cetmix_tower_server/models/cx_tower_key_mixin.py
+++ b/cetmix_tower_server/models/cx_tower_key_mixin.py
@@ -54,8 +54,15 @@ class CxTowerKeyMixin(models.AbstractModel):
             if key_parts:
                 key_refs.append(key_parts[1])
 
-        return key_model.search(
-            [
-                ("reference", "in", key_refs),
-            ]
-        )
+        return key_model.search(self._compose_secret_search_domain(key_refs))
+
+    def _compose_secret_search_domain(self, key_refs):
+        """Compose domain for searching secrets by references.
+
+        Args:
+            key_refs (List[str]): List of secret references.
+
+        Returns:
+            List: final domain for searching secrets.
+        """
+        return [("reference", "in", key_refs)]
--- a/cetmix_tower_server/README.rst
+++ b/cetmix_tower_server/README.rst
@@ -17,7 +17,7 @@ Cetmix Tower Server Management
     :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
     :alt: License: AGPL-3
 .. |badge3| image:: https://img.shields.io/badge/github-cetmix%2Fcetmix--tower-lightgray.png?logo=github
-    :target: https://github.com/cetmix/cetmix-tower/tree/14.0-dev/cetmix_tower_server
+    :target: https://github.com/cetmix/cetmix-tower/tree/14.0/cetmix_tower_server
     :alt: cetmix/cetmix-tower
 
 |badge1| |badge2| |badge3|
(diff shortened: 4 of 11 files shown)


cetmix_tower_server/models/cx_tower_key.py
```diff
@@ -50,11 +51,34 @@ class CxTowerKey(models.Model):
...
+                )
+                > 1
+            ):
+                raise ValidationError(
+                    _(
+                        "Reference must be unique for the combination of partner"
+                        " and server"
+                    )
+                )
+
```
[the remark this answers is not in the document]
ivs-cetmix: This SQL constraint will not work as expected because both partner_id and server_id fields can contain NULL values.

cetmix_tower_server/models/cx_tower_key.py
```diff
@@ -50,11 +51,34 @@ class CxTowerKey(models.Model):
     _sql_constraints = [
         (
             "reference_unique",
-            "UNIQUE(reference, partner_id, server_id)",
-            "Reference must be unique",
+            "check(1=1)",
```
GabbasovDinar: why not delete this constraint in migration script?

```sql
SELECT *
FROM information_schema.constraint_table_usage
WHERE table_name = 'cx_tower_key';
ALTER TABLE cx_tower_key DROP CONSTRAINT cx_tower_key_reference_unique;
```

P.S. not tested

GabbasovDinar (changes requested):

cetmix_tower_server/models/cx_tower_key.py
ivs-cetmix: This constraint is [defined](https://github.com/cetmix/cetmix-tower/blob/f62ffb248ba7e405e433f6b11d0f233581301946/cetmix_tower_server/models/cx_tower_reference_mixin.py#L29) at the mixin level, so it will be created by mixin. 
Probable solution can be deletion of this constraint in the post-init hook for new installations and in the migration script for existing ones.
@GabbasovDinar your thoughts?

tendil (approved):
Functional Review: LGTM

cetmix_tower_server/__init__.py
```diff
@@ -1,2 +1,87 @@
...
+            _logger.info(
+                f"Constraint '{constraint_name}' exists. " f"Proceeding with removal."
+            )
+
+            # Safely remove the constraint using parameterized query
+            cr.execute(
+                """
+                ALTER TABLE {} DROP CONSTRAINT {};
+            """.format(table_name, constraint_name)
+            )
```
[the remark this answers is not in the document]
tendil: #### **Parameterization Limitation in PostgreSQL**
Parameterized placeholders like `%s` **are not supported** for SQL identifiers, such as table names, column names, or constraint names.  
    - Placeholders `%s` can only be used to substitute **data values**, not structural elements of an SQL query.  
    - SQL identifiers (table names, constraint names) require direct, constant input because PostgreSQL does not allow dynamic injection of identifiers through placeholders.

cetmix_tower_server/migrations/14.0.0.4.1/post-migration.py
```diff
@@ -0,0 +1,80 @@
...
+            _logger.info(
+                f"Constraint '{constraint_name}' exists. Proceeding with removal."
+            )
+
+            # Safely remove the constraint using format-safe query
+            cr.execute(
+                """
+                ALTER TABLE {} DROP CONSTRAINT {};
+            """.format(table_name, constraint_name)
+            )
```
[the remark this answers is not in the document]
tendil: #### **Parameterization Limitation in PostgreSQL**
Parameterized placeholders like `%s` **are not supported** for SQL identifiers, such as table names, column names, or constraint names.  
    - Placeholders `%s` can only be used to substitute **data values**, not structural elements of an SQL query.  
    - SQL identifiers (table names, constraint names) require direct, constant input because PostgreSQL does not allow dynamic injection of identifiers through placeholders.

cetmix_tower_server/__init__.py
```diff
@@ -1,2 +1,87 @@
...
+    This ensures the database schema remains clean and consistent.
+    """
+    _logger.info("=== Starting `_post_init_hook`: Remove Unused Constraint ===")
+
+    table_name = "cx_tower_key"
+    constraint_name = "cx_tower_key_reference_unique"
+
+    try:
+        # Start transaction for atomic execution
+        cr.execute("BEGIN;")
```
ivs-cetmix: Do we need to start a transaction explicitly? Would it be started by default?

tendil: 1. Yes, the transaction is handled automatically by Odoo itself, so the explicit BEGIN; could have been omitted. I added it just in case, to explicitly show that the transaction is going to start, and to make it easier to read for those who are going to work with this code. But there is usually no need.

2. On test - it is still running by default as we are overriding it. I'll delete it.

cetmix_tower_server/migrations/14.0.0.4.1/post-migration.py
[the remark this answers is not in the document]
tendil: Снимок экрана 2024-12-18 в 18 07 57

::

**Error description**:  
Parameterised placeholders `%s` for SQL identifiers (such as table names or constraints) are **not supported** in PostgreSQL. If you use parameterisation for `ALTER TABLE', PostgreSQL treats the names as **string literals** and encloses them in single quotes, resulting in a syntax error.  

**Why it doesn't work**:  
```Python
cr.execute("ALTER TABLE %s DROP CONSTRAINT %s;", (table_name, constraint_name))
```
The parameters `table_name' and `constraint_name' are passed as `'cx_tower_key' and `'reference_unique' strings, but in SQL syntax, table and constraint names **don't need to be enclosed in quotes**, unlike data values.

**Why I use `.format`:  
I use `.format` (or f-strings) to replace table and constraint names dynamically. This is required as PostgreSQL allows no parameterisation of SQL identifiers. I add a check before executing the query to ensure that the values are valid identifiers.  

**Summary**:  
For table names and constraints, parameterisation with `%s` does not work. Therefore, to avoid SQL injection and errors, **`.format`** is used with additional validation of values for correctness.

cetmix_tower_server/__init__.py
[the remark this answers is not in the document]
tendil: Same here.

cetmix_tower_server/__init__.py
```diff
@@ -1,2 +1,83 @@
...
+
+import logging
+
 from . import models
 from . import wizards
+
+_logger = logging.getLogger(__name__)
+
+
+def _post_init_hook(cr, registry):
```
GabbasovDinar: Post_init_hook are the functions that will execute after the module installation. Maybe in your case you need to use migrations to work with the exists data

GabbasovDinar (changes requested):

cetmix_tower_server/__init__.py
tendil: I was given two things to do as part of this task:
1) For new installations: add a post-init hook that deletes constraint from the database.
2) For existing installations: add a migration script that does the same.

This is what has been implemented in my solution, namely the post_init_hook and the migration.

cetmix_tower_server/models/cx_tower_key_mixin.py
```diff
@@ -54,8 +54,15 @@ def _extract_secret_ids(self, code):
...
+    def _compose_secret_search_domain(self, key_refs):
+        """Compose domain for searching secrets by references.
+
+        Args:
+            key_refs (List[str]): List of secret references.
+
+        Returns:
+            List: final domain for searching secrets.
+        """
+        return [("reference", "in", key_refs)]
```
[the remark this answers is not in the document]
tendil: This is not relevant to our case

Aldeigja (approved):
Functional LGTM

cetmix_tower_server/__init__.py
tendil: @GabbasovDinar 

This is what has been implemented in my solution, namely the post_init_hook and the migration.

GabbasovDinar: @tendil why do you resolve a dialog without a response that is also directed to the author of the PR?

GabbasovDinar (approved):

CetmixGitDrone:

cetmix_tower_server/__init__.py
tendil: @GabbasovDinar My mistake, please accept my apologies. I made that comment Unresolved.

CetmixGitDrone:
On my way to merge this fine PR!
Prepared branch [14.0-dev-ocabot-merge-pr-158-by-ivs-cetmix-bump-nobump](https://github.com/cetmix/cetmix-tower/commits/14.0-dev-ocabot-merge-pr-158-by-ivs-cetmix-bump-nobump), awaiting test results.

CetmixGitDrone:
Congratulations, your PR was merged at 596ed67b4f0f17ed674fb78085bb9ec2f3f9a532. Thanks a lot for contributing to cetmix. ❤️

PR #182, opened 2025-01-14
[IMP] cetmix_tower_server: Access to server fields

GabbasovDinar:
Limit access to the following fields of the `cx.tower.server` ​model to Manager and Root:
- OS
- IP v4 Address
- IP v6 Address
- SSH username 
- SSH Port
- SSS auth mode
- SSH key
- Use sudo

Task: 4116

## Summary by CodeRabbit

- **Security Enhancements**
	- Added group-based access control for server connection settings.
	- Restricted sensitive server configuration fields to the manager group.
	- Made SSH password field optional.

- **User Interface**
	- Introduced a new kanban view for the manager group with restricted field access.
	- Updated search view to reflect new access controls and visibility based on user roles.

- **Documentation**
	- Added note about user access levels in the configuration guide.

- **Testing**
	- Enhanced test coverage for user access and command execution, including sensitive variables.
	- Removed a test for unique constraint validation on variable options.
	- Streamlined test cases by removing unnecessary mocks and directly testing functionality.
	- Added new tests for verifying user access to files with sensitive variables.

### DIFF
--- a/cetmix_tower_server/models/cx_tower_file.py
+++ b/cetmix_tower_server/models/cx_tower_file.py
@@ -43,6 +43,7 @@ class CxTowerFile(models.Model):
     name = fields.Char(help="File name WITHOUT path. Eg 'test.txt'")
     rendered_name = fields.Char(
         compute="_compute_render",
+        compute_sudo=True,
     )
     template_id = fields.Many2one(
         "cx.tower.file.template",
@@ -57,6 +58,7 @@ class CxTowerFile(models.Model):
     )
     rendered_server_dir = fields.Char(
         compute="_compute_render",
+        compute_sudo=True,
     )
     full_server_path = fields.Char(
         compute="_compute_full_server_path",
@@ -117,6 +119,7 @@ class CxTowerFile(models.Model):
     )
     rendered_code = fields.Char(
         compute="_compute_render",
+        compute_sudo=True,
         help="File content with variables rendered",
     )
     keep_when_deleted = fields.Boolean(
@@ -644,7 +647,7 @@ class CxTowerFile(models.Model):
                     )
                 else:
                     return False
-                file.server_response = "ok"
+                file.sudo().server_response = "ok"
             except Exception as error:
                 if raise_error:
                     raise ValidationError(
@@ -698,4 +701,4 @@ class CxTowerFile(models.Model):
                 )
             if file.server_response == "ok":
                 vals.update({"sync_date_last": last_sync_date})
-            file.write(vals)
+            file.sudo().write(vals)
--- a/cetmix_tower_server/models/cx_tower_server.py
+++ b/cetmix_tower_server/models/cx_tower_server.py
@@ -280,15 +280,29 @@ class CxTowerServer(models.Model):
     )
 
     # ---- Connection
-    ip_v4_address = fields.Char(string="IPv4 Address")
-    ip_v6_address = fields.Char(string="IPv6 Address")
-    ssh_port = fields.Char(string="SSH port", required=True, default="22")
-    ssh_username = fields.Char(string="SSH Username", required=True)
-    ssh_password = fields.Char(string="SSH Password")
+    ip_v4_address = fields.Char(
+        string="IPv4 Address", groups="cetmix_tower_server.group_manager"
+    )
+    ip_v6_address = fields.Char(
+        string="IPv6 Address", groups="cetmix_tower_server.group_manager"
+    )
+    ssh_port = fields.Char(
+        string="SSH port",
+        required=True,
+        default="22",
+        groups="cetmix_tower_server.group_manager",
+    )
+    ssh_username = fields.Char(
+        string="SSH Username", required=True, groups="cetmix_tower_server.group_manager"
+    )
+    ssh_password = fields.Char(
+        string="SSH Password", groups="cetmix_tower_server.group_manager"
+    )
     ssh_key_id = fields.Many2one(
         comodel_name="cx.tower.key",
         string="SSH Private Key",
         domain=[("key_type", "=", "k")],
+        groups="cetmix_tower_server.group_manager",
     )
     ssh_auth_mode = fields.Selection(
         string="SSH Auth Mode",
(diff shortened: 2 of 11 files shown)


cetmix_tower_server/models/cx_tower_server.py
```diff
@@ -279,15 +279,29 @@ class CxTowerServer(models.Model):
...
+    )
+    ip_v6_address = fields.Char(
+        string="IPv6 Address", groups="cetmix_tower_server.group_manager"
+    )
+    ssh_port = fields.Char(
+        string="SSH port",
+        required=True,
+        default="22",
+        groups="cetmix_tower_server.group_manager",
+    )
```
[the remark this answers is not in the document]
GabbasovDinar: user cannot create records for this model, only read, so it's ok

cetmix_tower_server/models/cx_tower_server.py
```diff
@@ -279,15 +279,29 @@
...
+    ssh_username = fields.Char(
+        string="SSH Username", required=True, groups="cetmix_tower_server.group_manager"
+    )
+    ssh_password = fields.Char(
+        string="SSH Password", groups="cetmix_tower_server.group_manager"
+    )
```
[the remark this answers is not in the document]
GabbasovDinar: user cannot create records for this model, only read, so it's ok

cetmix_tower_server/views/cx_tower_server_view.xml
```diff
@@ -360,7 +372,7 @@
                 <field name="name" />
                 <field name="reference" />
                 <field name="status" />
-                <field name="os_id" />
+                <field name="os_id" groups="cetmix_tower_server.group_manager" />
```
ivs-cetmix: Question: why do we need to add `group` attribute to the view definition?

ivs-cetmix (changes requested):
We have changed the flow so need to update the test coverage accordingly.
Eg we need to ensure that user with group "Tower/User" can still run SSH commands with no issues.

cetmix_tower_server/views/cx_tower_server_view.xml
GabbasovDinar: I agree, I will fix it

cetmix_tower_server/tests/test_command.py
```diff
@@ -529,6 +529,16 @@ def test_user_access_rule(self):
         # Ensure that user can access the command
         command_name = test_command_1_as_bob.name
         self.assertEqual(command_name, "Test command", msg="Must return 'Test command'")
+        # Check that user with "cetmix_tower_server.group_user" can execute ssh command
+        test_command.write(
```
ivs-cetmix, line 540: I think we need to update [this method](https://github.com/cetmix/cetmix-tower/blob/c428b9e07d021f38d09889d7cf898b4c9d005d3c/cetmix_tower_server/tests/common.py#L219) too. And add some read operations to ensure that they don't trigger any access issues.
However let's merge https://github.com/cetmix/cetmix-tower/pull/181 first

ivs-cetmix (changes requested):
I think we need

cetmix_tower_server/wizards/cx_tower_command_execute_wizard.py
```diff
@@ -53,6 +53,7 @@ class CxTowerCommandExecuteWizard(models.TransientModel):
     any_server = fields.Boolean(default=True)
     rendered_code = fields.Text(
         compute="_compute_rendered_code",
+        compute_sudo=True,
```
ivs-cetmix, line 56: Shouldn't we add `compute_sudo` [here](https://github.com/cetmix/cetmix-tower/blob/3c3f48a4767ca4d83b97913fa3de7914db01aa28/cetmix_tower_server/models/cx_tower_file.py#L119) too?

ivs-cetmix (changes requested):

Aldeigja (approved):
functional LGTM

cetmix_tower_server/wizards/cx_tower_command_execute_wizard.py
GabbasovDinar: yes, thanks!

cetmix_tower_server/models/cx_tower_file.py
```diff
@@ -644,7 +647,7 @@ def _process(self, action, raise_error=False):
                     )
                 else:
                     return False
-                file.server_response = "ok"
+                file.sudo().server_response = "ok"
```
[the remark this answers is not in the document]
GabbasovDinar: can be a some command that allows user to start loading a file, by default the user does not have access to modify the file, but the command must be executed, so this method allows you to update techical data on the file when the command is executed.

cetmix_tower_server/tests/test_file.py
```diff
@@ -392,3 +392,33 @@ def test_file_with_secret_key(self):
...
+                "cetmix_tower_server.group_root",
+            ],
+        )
+        # Add bob to user group
+        self.add_to_group(self.user_bob, "cetmix_tower_server.group_user")
+        # Add bob as subscriber of the server to allow upload file
+        self.server_test_1.message_subscribe([self.user_bob.partner_id.id])
+        # Upload file to server
+        file.with_user(self.user_bob).action_push_to_server()
+        self.assertEqual(file.server_response, "ok")
```
ivs-cetmix, line 382: What if `server_response` was already "ok" before we pushed the file? And our push operation simply didn't check it.
Probably we need to set `server_response` to some other value before running the test.

ivs-cetmix (changes requested):
Note: no need to add `refracting` in the commit message again because there is a [REF] tag already. So something like:

```
[REF] cetmix_tower_server: Tests

Use MagicMock for mocking SSH and file operations.
This is done to make the test process more clean and clear.
```

cetmix_tower_server/tests/test_file.py
GabbasovDinar: done

ivs-cetmix (approved):
Code review LGTM

CetmixGitDrone:

hoangtrann (approved):
Code LGTM

CetmixGitDrone:
This PR looks fantastic, let's merge it!
Prepared branch [14.0-dev-ocabot-merge-pr-182-by-ivs-cetmix-bump-nobump](https://github.com/cetmix/cetmix-tower/commits/14.0-dev-ocabot-merge-pr-182-by-ivs-cetmix-bump-nobump), awaiting test results.

CetmixGitDrone:
Congratulations, your PR was merged at 9e8bc1fdb5ec4a3eb5bc8bfef0204500be029b3f. Thanks a lot for contributing to cetmix. ❤️

PR #166, opened 2024-12-23
[REF][IMP] Access Rules for Variables and Values with Tests

tendil:
### This PR is irrelevant, as it was having problems versioning the Branch. As part of this task, a [new PR has been opened for it](https://github.com/cetmix/cetmix-tower/pull/187)

---

[REF]
- Refactored existing code to align with PEP8 and Odoo guidelines.
- Improved indentation and formatting in affected files for better readability.

[IMP]
- Added access control by inheriting `cx.tower.access.mixin` in `TowerVariable` and `TowerVariableValue` models.
- Introduced the `access_level` field in the models and views, allowing dynamic control of variable visibility.
- Updated search and form views to include the `access_level` field for better usability.

[NEW]
- Created a new security file `cx_tower_variable_security.xml` to define access rules for `TowerVariable`.
- Updated `cx_tower_variable_value_security.xml` to include refined domain rules for `TowerVariableValue`.

[TEST]
- Updated test `test_variable_value_access` to validate refined access logic.
- Added a new test `test_variable_value_server_subscription` to ensure server subscription-based variable visibility.

Steps to Reproduce
1. Create a new variable and assign it to a server with specific `access_level`.
2. Test visibility of the variable based on user roles (`group_user`, `group_manager`, `group_root`).
3. Validate global and local variable toggling and their visibility to respective roles.
4. Subscribe a user to the server and ensure variable visibility updates dynamically.
5. Run the tests (`test_variable_value_access`, `test_variable_value_server_subscription`) to confirm correct behavior.

File Changes
1. `cx_tower_variable_security.xml`
   - Created a new security file to manage access rules for `TowerVariable`.

2. `cx_tower_variable_value_security.xml`
   - Updated domain rules to dynamically control visibility based on `access_level` and server subscriptions.

3. Models (`TowerVariable`, `TowerVariableValue`)
   - Inherited `cx.tower.access.mixin`.
   - Added `access_level` fields and their logic.

4. Views (`TowerVariable`, `TowerVariableValue`)
   - Updated forms and search views to include `access_level` fields.

5. Tests
   - Refined `test_variable_value_access` to ensure compatibility with new logic.
   - Added `test_variable_value_server_subscription` to validate subscription-based access.

6. General
   - Minor corrections in formatting, indentation, and adherence to Odoo development guidelines.

Task: 4055

## Release Notes

- **New Features**
  - Enhanced access control for Tower Variables and Variable Values.
  - Added granular access levels for users, managers, and root users.
[...]

PR #187, opened 2025-01-17
[IMP][REF][ADD] cetmix_tower_server: Refactor Access Rules and Add Te…

tendil:
…sts for Variables and Values

---

### So as not to lose any old comments and edits, and to keep them in the history - have a [look at the old PR](https://github.com/cetmix/cetmix-tower/pull/166).

---

This PR improves access control logic, introduces new features, and refines existing functionality for `TowerVariable` and `TowerVariableValue` models. The changes align with OCA guidelines and include comprehensive test coverage.

Refactoring:
- Refactored code to follow PEP8 and Odoo development guidelines.
- Improved readability with better indentation and formatting.

Improvements:
- Enhanced access control by inheriting `cx.tower.access.mixin` in `TowerVariable` and `TowerVariableValue`.
- Added `access_level` field in models and views for dynamic control of variable visibility.
- Updated search and form views to include `access_level`.

Additions:
- Created `cx_tower_variable_security.xml` to define access rules for `TowerVariable`.
- Refined `cx_tower_variable_value_security.xml` to include updated domain rules for `TowerVariableValue`.

Tests:
- Added four new tests:
  - `test_variable_access_levels`: Verifies access based on `access_level`.
  - `test_variable_access_rules`: Ensures correct rules for variables.
  - `test_variable_value_access_rules`: Validates refined access logic.
  - `test_variable_rendering_in_execution`: Confirms all variables are rendered correctly during execution.
- Updated the existing test `test_variable_value_access` to ensure compatibility with new logic.

Steps to Reproduce:
1. Create a variable and assign it to a server with specific `access_level`.
2. Test variable visibility based on roles (`group_user`, `group_manager`, `group_root`).
3. Validate global and local variable toggling and their visibility to respective roles.
4. Subscribe a user to a server and ensure visibility updates dynamically.
5. Run tests to confirm functionality.

File Changes:
1. `cx_tower_variable_security.xml`: New file to manage access rules for `TowerVariable`.
2. `cx_tower_variable_value_security.xml`: Updated domain rules for `TowerVariableValue`.
3. Models (`TowerVariable`, `TowerVariableValue`):
   - Inherited `cx.tower.access.mixin`.
   - Added `access_level` fields with required logic.
4. Views:
   - Updated forms and search views to include `access_level`.
5. Tests:
   - Refined `test_variable_value_access`.
   - Added `test_variable_value_server_subscription`.

Task: 4055

## Summary by CodeRabbit

- **Documentation**
	- Updated user documentation with stable branch URLs, refreshed images, and improved formatting.
- **New Features**
	- Introduced enhanced access control with an “access level” attribute across variable and server configurations.
	- Expanded search and display views, improving role-based management.
- **Chore**
	- Updated the module version to the latest release.
- **Security**
	- Refined access rules to clearly delineate rights for different user roles.


cetmix_tower_server/static/description/index.html
```diff
@@ -1131,16 +1124,16 @@ <h1>Usage</h1>
 <div class="section" id="create-a-new-server-from-a-server-template">
 <h2>Create a new Server from a Server Template</h2>
 <ul class="simple">
-<li>Go to the <tt class="docutils literal">Cetmix Tower/Servers/Templates</tt> menu and select a
-<a class="reference external" href="CONFIGURE.md/#configure-a-server-template">Server Template</a></li>
+<li>Go to the <tt class="docutils literal">Cetmix Tower/Servers/Templates</tt> menu and select a <a class="reference external" href="CONFIGURE.md/#configure-a-server-template">Server
+Template</a></li>
```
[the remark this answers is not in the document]
tendil: This file is generated automatically.

cetmix_tower_server/models/cx_tower_variable_value.py
```diff
@@ -144,6 +145,15 @@ def _onchange_variable_id(self):
             else:
                 rec.option_id = None
 
+    @api.depends("variable_id")
+    def _compute_access_level(self):
```
ivs-cetmix, line 143: [This field](https://github.com/cetmix/cetmix-tower/blob/4a45049f9367c3ad80188e1daf1382ff7c095b2b/cetmix_tower_server/models/cx_tower_access_mixin.py#L35) doesn't have compute method. How do you suppose this function to be triggered?

cetmix_tower_server/views/cx_tower_server_view.xml
```diff
@@ -266,7 +266,7 @@
                         <page
                             name="configuration"
                             string="Configuration"
-                            groups="cetmix_tower_server.group_manager"
+                            groups="cetmix_tower_server.group_manager, cetmix_tower_server.group_user"
```
ivs-cetmix: Just remove the "groups" key.

ivs-cetmix (changes requested):

ivs-cetmix:
Supersedes https://github.com/cetmix/cetmix-tower/pull/166

cetmix_tower_server/models/cx_tower_variable_value.py
tendil: But if we make the [This field](https://github.com/cetmix/cetmix-tower/blob/4a45049f9367c3ad80188e1daf1382ff7c095b2b/cetmix_tower_server/models/cx_tower_access_mixin.py#L35) field calculable, then we change this logic to all the given places where this field is used.
What is the best way to proceed in this case?

Or maybe it would be better to make this field calculable, but only for the `Configuration` tab context in `Server`?

ivs-cetmix: You should make it computed only for this particular model.

CetmixGitDrone:

cetmix_tower_server/static/description/index.html
```diff
@@ -369,7 +369,7 @@ <h1 class="title">Cetmix Tower Server Management</h1>
...
+<p><a class="reference external image-reference" href="https://odoo-community.org/page/development-status"><img alt="Beta" src="https://img.shields.io/badge/maturity-Beta-yellow.png" /></a> <a class="reference external image-reference" href="http://www.gnu.org/licenses/agpl-3.0-standalone.html"><img alt="License: AGPL-3" src="https://img.shields.io/badge/licence-AGPL--3-blue.png" /></a> <a class="reference external image-reference" href="https://github.com/cetmix/cetmix-tower/tree/14.0/cetmix_tower_server"><img alt="cetmix/cetmix-tower" src="https://img.shields.io/badge/github-cetmix%2Fcetmix--tower-lightgray.png?logo=github" /></a></p>
```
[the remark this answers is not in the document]
tendil: We don't need to change anything in this file by hand because it is generated automatically.

cetmix_tower_server/static/description/index.html
```diff
@@ -1132,16 +1125,16 @@
 <div class="section" id="create-a-new-server-from-a-server-template">
 <h2>Create a new Server from a Server Template</h2>
 <ul class="simple">
-<li>Go to the <tt class="docutils literal">Cetmix Tower/Servers/Templates</tt> menu and select a
-<a class="reference external" href="CONFIGURE.md/#configure-a-server-template">Server Template</a></li>
+<li>Go to the <tt class="docutils literal">Cetmix Tower/Servers/Templates</tt> menu and select a <a class="reference external" href="CONFIGURE.md/#configure-a-server-template">Server
+Template</a></li>
```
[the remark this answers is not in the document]
tendil: We don't need to change anything in this file by hand because it is generated automatically.

cetmix_tower_server/tests/test_plan.py
```diff
@@ -1583,6 +1583,8 @@ def test_plan_line_action_variable_values_user_access_rights(self):
         #  plan line action
         # Add user to group
         self.add_to_group(self.user_bob, "cetmix_tower_server.group_user")
+        self.variable_os.write({"access_level": "1"})
```
ivs-cetmix: You should use the helper [write_and_invalidate](https://github.com/cetmix/cetmix-tower/blob/dbceb79bd45e80a0436632f2c383182619eda61b/cetmix_tower_server/tests/common.py#L362) method to ensure that record is updated properly. Check how it used in other places in this file, eg [here](https://github.com/cetmix/cetmix-tower/blob/dbceb79bd45e80a0436632f2c383182619eda61b/cetmix_tower_server/tests/test_plan.py#L1601).

cetmix_tower_server/models/cx_tower_variable_value.py
```diff
@@ -144,6 +150,16 @@ def _onchange_variable_id(self):
             else:
                 rec.option_id = None
 
+    @api.depends("variable_id")
```
ivs-cetmix: What if variable access level has changed? Eg it was "2" and then became "1". This means "User" group members will not be able to access the variable values although they should.

cetmix_tower_server/tests/test_plan.py
```diff
@@ -1583,6 +1583,8 @@ def test_plan_line_action_variable_values_user_access_rights(self):
...
+        self.variable_value.write({"access_level": "1"})
         test_plan_line_action_as_bob = self.plan_3_line_1_action.with_user(
```
ivs-cetmix, line 1587: This is a computed field that should be updated automatically when the variable access level is updated.

ivs-cetmix (changes requested):
Please do not use AI to generate the PR description. Write it in your own words, otherwise it's not clear if you really understand what is done in the code.

tendil:
> Please do not use AI to generate the PR description. Write it in your own words, otherwise it's not clear if you really understand what is done in the code.

---

At the `Prepare For Merge` stage I update it to the level of what is implemented and in order to get a complete description I prescribe everything myself and then correct it with the AI.

cetmix_tower_server/views/cx_tower_server_view.xml
```diff
@@ -306,6 +302,8 @@
                                         options="{'no_create': True, 'no_create_edit': True}"
                                         attrs="{'readonly': [('variable_type', '=', 's')]}"
                                     />
+                                    <field name="access_level" />
+                                    <field name="option_ids_domain" invisible="1" />
```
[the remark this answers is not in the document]
tendil: Outdated comments. This error was caused by an incorrect rebase. I've fixed it now

cetmix_tower_server/models/cx_tower_variable.py
```diff
@@ -40,6 +40,14 @@ def _compute_value_ids_count(self):
         for rec in self:
             rec.value_ids_count = len(rec.value_ids)
 
+    @api.onchange("access_level")
+    def _onchange_access_level(self):
```
ivs-cetmix: What if access level is updated from code? And you already have a "compute" method in the variable values. Why do we need another function for the same thing?

ivs-cetmix (changes requested):

cetmix_tower_server/models/cx_tower_variable.py
tendil: The only other option I have left is to make - `inverse="_inverse_access_level"` field to correctly change the access level for variable_id as well.

cetmix_tower_server/models/cx_tower_variable_value.py
```diff
@@ -133,6 +140,25 @@ def _onchange_variable_id(self):
             if rec.variable_id.option_ids:
                 rec.value_char = False
 
+    @api.depends("variable_id.access_level")
```
ivs-cetmix: What if variable has changed?

cetmix_tower_server/models/cx_tower_variable_value.py
```diff
@@ -133,6 +140,25 @@ def _onchange_variable_id(self):
...
+    def _compute_access_level(self):
+        """
+        Automatically set the `access_level` based on `variable_id.access_level`
+        when adding a variable to the server.
+        """
+        for rec in self:
+            if rec.variable_id:
+                rec.access_level = rec.variable_id.access_level
+
+    def _inverse_access_level(self):
```
ivs-cetmix: What is this method needed for? Is there are need to change the variable access level if variable value access level is changed?
And it there was such a request, why not to use a related field ?

ivs-cetmix (changes requested):

cetmix_tower_server/models/cx_tower_variable_value.py
tendil: Replace:
`@api.depends("variable_id.access_level")` --> `@api.depends("variable_id", "variable_id.access_level")`

cetmix_tower_server/models/cx_tower_variable_value.py
tendil: Related is **not suitable** because:
 - related=`variable_id.access_level` automatically changes the access level of variable when `variable.value.access_level` is changed.
 - This breaks the logic, as the user should only control the level of variable.value, and it should not affect the variable itself.

---

So I revised the logic and redesigned the _inverse_access_level method so that it also replaces the old `_check_access_level_consistency` method.

cetmix_tower_server/models/cx_tower_variable_value.py
```diff
@@ -169,6 +186,30 @@ def _compute_variable_ids(self):
                 ["value_char"], force_record=record
             )
 
+    def _inverse_access_level(self):
```
ivs-cetmix: You are not using this function to set a value, just to check it. Why do you need the inverse method and not to use `@api.constraints` instead?

cetmix_tower_server/models/cx_tower_variable_value.py
```diff
@@ -78,6 +79,12 @@ class TowerVariableValue(models.Model):
         store=True,
     )
     required = fields.Boolean()
+    access_level = fields.Selection(
+        compute="_compute_access_level",
+        inverse="_inverse_access_level",
```
ivs-cetmix: This field is writable (readonly=False). What is the `inverse` method used for?

ivs-cetmix (changes requested):

cetmix_tower_server/models/cx_tower_variable_value.py
tendil: Initially, we also considered using `@api.constrains` instead of the inverse method. However, the key reason for using `_inverse_access_level` in this case is that we need to enforce consistency between `variable_value.access_level` and `variable.access_level` when users manually edit the value.

Without the inverse method, users could manually change access_level of variable_value without triggering automatic updates.
This could lead to cases where the variable’s level is 2, but the `value level` is 1, violating our constraints.

Thus, the inverse method is needed to ensure bidirectional consistency between `variable_value.access_level` and `variable.access_level`. Simply using `@api.constrains` would not allow us to properly enforce updates when the `value` is manually changed.

cetmix_tower_server/models/cx_tower_variable_value.py
tendil: There's also.

---

- The inverse method ensures that manual edits to access_level behave correctly.
- It prevents inconsistencies where the value is manually changed but not validated.
- Without it, computed values wouldn't update properly when manually edited.

cetmix_tower_server/models/cx_tower_variable_value.py
ivs-cetmix: Current implementation of the `inverse` method doesn't update any field. The only thing it does is that it checks the values and raises an exception. So there is absolutely no difference from the `api.constaints` method. 
If you change the decorator the behaviour will remain exactly the same.

ivs-cetmix (changes requested):

ivs-cetmix (changes requested):
Nothing has changed since the previous review.

ivs-cetmix (approved):
LGTM for now

ivs-cetmix:
Done in https://github.com/cetmix/cetmix-tower/pull/200

PR #192, opened 2025-01-24
[FIX] Update head field in CxTowerGitRemote model to be optional

GSLabIt:
## Summary by CodeRabbit

- **New Features**
	- Updated Git remote head field to be optional
	- Added help text for Git remote head configuration

- **Improvements**
	- Enhanced flexibility in specifying Git remote head settings
	- Allows auto-detection when head is not explicitly defined

### DIFF
--- a/cetmix_tower_git/models/cx_tower_git_remote.py
+++ b/cetmix_tower_git/models/cx_tower_git_remote.py
@@ -50,7 +50,6 @@ class CxTowerGitRemote(models.Model):
         readonly=False,
     )
     head = fields.Char(
-        required=True,
         help="Git remote head. Link to branch, PR, commit or commit hash."
         " Leave blank to auto-detect",
     )


ivs-cetmix:
Hello @GSLabIt , thank you for your contribution! Could you please provide some specific example of when a remote head can be empty?

GSLabIt:
> Hello @GSLabIt , thank you for your contribution! Could you please provide some specific example of when a remote head can be empty?

Not really sure, but if u want to allow `"Leave blank to auto-detect"`,  field must be not `required` 😅

ivs-cetmix:
> > Hello @GSLabIt , thank you for your contribution! Could you please provide some specific example of when a remote head can be empty?
> 
> Not really sure, but if u want to allow `"Leave blank to auto-detect"`, field must be not `required` 😅

I have created the following file to test if head can be left blank:

```test.yaml
./project:
    remotes:
        oca: https://github.com/oca/project.git
    merges:
        -
            remote: oca
```

Got the following error trying to clone it:

```bash
git_aggregator.exception.ConfigException: /Users/ivansokolov/odoo/tmp-addons/project: Merge lacks mandatory `remote` or `ref` keys.
```

Taking into account, that Git Aggregator is currently the default Git tool used with Git Project, making the `head` field non-mandatory might lead to an unpredictable behaviour.
Or it will require to implement some complex checks and controls, eg remote cannot be set as "enabled" unless there is a head provided for it.
So I don't think it's reasonable to make this field non required in the base module.
However if you need this field to be non-required for some further customisation I suggest simply overriding the `required` key in your module.

PR #194, opened 2025-01-28
[IMP] cetmix_tower_server: Access rules

GabbasovDinar:
Added new security rules for tower key management

Task: 4300

## Summary by CodeRabbit

- **Security Enhancements**
	- Added a new security-related XML file for key management.
	- Implemented access control rules for the `cx_tower_key` model based on user groups and permissions.

- **Data Model Updates**
	- Introduced `server_ids` field to link partners with servers.
	- Expanded model relationships to improve data associations.

- **Testing Improvements**
	- Enhanced test coverage for user access rules.
	- Added comprehensive tests for command and key access permissions.

- **Documentation Updates**
	- Updated URLs and formatting in the README and HTML documentation to reflect version changes.

### DIFF
--- a/cetmix_tower_server/models/res_partner.py
+++ b/cetmix_tower_server/models/res_partner.py
@@ -1,0 +1,16 @@
+# Copyright (C) 2025 Cetmix OÜ
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+
+from odoo import fields, models
+
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    server_ids = fields.One2many(
+        string="Servers",
+        comodel_name="cx.tower.server",
+        inverse_name="partner_id",
+        help="Cetmix Tower servers that belong to this partner",
+    )
--- a/cetmix_tower_server/tests/test_command.py
+++ b/cetmix_tower_server/tests/test_command.py
@@ -645,6 +645,49 @@ COMMAND_RESULT = {
             msg="Command name should be same",
         )
 
+    def test_user_access_rule_with_keys(self):
+        """Test user access rule"""
+
+        # create server key
+        server_key = self.Key.create(
+            {
+                "name": "server key",
+                "secret_value": "server key value",
+                "key_type": "s",
+                "server_id": self.server_test_1.id,
+            }
+        )
+
+        # Create the test command with server key
+        code = f"mkdir {server_key.reference_code}"
+        command_with_key = self.Command.create(
+            {"name": "Command with key", "code": code, "access_level": "1"}
+        )
+
+        # Remove bob from all cxtower_server groups
+        self.remove_from_group(
+            self.user_bob,
+            [
+                "cetmix_tower_server.group_user",
+                "cetmix_tower_server.group_manager",
+                "cetmix_tower_server.group_root",
+            ],
+        )
+
+        # Add user to group
+        self.add_to_group(self.user_bob, "cetmix_tower_server.group_user")
+        # User can execute command with key
+        self.server_test_1.with_user(self.user_bob).execute_command(
+            command_with_key,
+        )
+
+        # Add user to group_manager
+        self.add_to_group(self.user_bob, "cetmix_tower_server.group_manager")
+        # Manager can execute command with key
+        self.server_test_1.with_user(self.user_bob).execute_command(
+            command_with_key,
+        )
+
     def test_parse_ssh_command_result(self):
         """Test ssh command result parsing"""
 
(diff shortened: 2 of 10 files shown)


cetmix_tower_server/models/res_partner.py
```diff
@@ -0,0 +1,15 @@
...
+from odoo import fields, models
+
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    server_ids = fields.One2many(
+        string="Servers",
+        comodel_name="cx.tower.server",
+        inverse_name="partner_id",
```
ivs-cetmix, line 14: help="Cetmix Tower servers that belong to this partner"

cetmix_tower_server/tests/test_key.py
ivs-cetmix, line 1: You have added new access rules. This means you must add access rules to check those new rules to ensure they are working properly.

ivs-cetmix (changes requested):
You have added new access rules. This means you must add access rules to check those new rules to ensure they are working properly.
There must be test for:
1. Keys. Ensure that access rules are respected.
2. Commands: Ensure that new rules don't break the command flow. 

User may not have access to a secret, but if he runs a command this secret still needed to be rendered properly. 
Eg If someone with the "User" access lever runs a Flight Plan and there is a command that has a secret in it, all 3 steps of the secret resolution must pass:
- Server secret
- Partner secret
- Global secret

cetmix_tower_server/tests/test_command.py
```diff
@@ -644,6 +645,49 @@ def test_user_access_rule(self):
...
 
+    def test_user_access_rule_with_keys(self):
+        """Test user access rule"""
+
+        # create server key
+        server_key = self.Key.create(
+            {
+                "name": "server key",
+                "secret_value": "server key value",
+                "key_type": "s",
```
ivs-cetmix, line 656: I would suggest to add access level explicitly in order to simplify possible error investigation in case default access level is somehow changed, eg in another module.

cetmix_tower_server/tests/test_key.py
```diff
@@ -27,33 +27,24 @@ def test_key_creation(self):
...
         # Add user to group
         self.add_to_group(self.user_bob, "cetmix_tower_server.group_user")
 
-        # Get value
-        key_value = key_bob.secret_value
-
-        # Ensure placeholder is used instead of the key value
-        self.assertEqual(
-            key_value,
-            SECRET_VALUE_PLACEHOLDER,
```
ivs-cetmix, line 54: We need to ensure that SECRET_VALUE_PLACEHOLDER is returned properly., so this test cannot be removed.
If you can modify it or add another one if this one is not working in the new flow.

cetmix_tower_server/tests/test_key.py
```diff
@@ -72,10 +63,97 @@ def test_key_access_rights(self):
...
+
+        # Manager cannot read key value
+        with self.assertRaises(AccessError):
+            key_bob.read(["secret_value"])
+
+        # Add Bob as subscriber to server
+        self.server_test_1.message_subscribe([self.user_bob.partner_id.id])
+
+        # Now manager server subscriber can read and update this key
+        key_bob.read(["secret_value"])
```
ivs-cetmix, line 113: But which value does he get? Plain text password, or a placeholder?
Should't we try to write first before reading, so we could ensure that the secret values is properly masked?

cetmix_tower_server/tests/test_key.py
```diff
@@ -72,10 +63,97 @@ def test_key_access_rights(self):
...
+        # Add Bob to Manager group
+        self.add_to_group(self.user_bob, "cetmix_tower_server.group_manager")
+
+        # Ensure that manager don't has access to key
+        with self.assertRaises(AccessError):
+            # Get value
+            key_bob.read(["secret_value"])
+
+        # Ensure that key has no servers
+        self.assertFalse(self.key_1.server_id)
```
ivs-cetmix, line 147: Shouldn't this check be done at the very beginning of the test?

cetmix_tower_server/tests/test_key.py
```diff
@@ -72,10 +63,97 @@ def test_key_access_rights(self):
...
+        self.assertFalse(self.key_1.server_ssh_ids)
+
+        # Add Bob as subscriber to server
+        partner.server_ids.message_subscribe([self.user_bob.partner_id.id])
+
+        # Now manager server subscriber can read and update this key
+        key_bob.read(["secret_value"])
```
ivs-cetmix, line 154: What is the value? Should't we try to write first before reading, so we could ensure that the secret values is properly masked?

ivs-cetmix (changes requested):

ivs-cetmix:
Done in https://github.com/cetmix/cetmix-tower/pull/203

PR #222, opened 2025-03-11
[14.0][IMP] cetmix_tower_server: use checkbox for sudo mode

halbtonjazz:
Replace the sudo selector with a checkbox in the command wizard to ensure the sudo mode is determined by the server settings. This change avoids mismatches between the user selection and server configuration by validating the sudo mode before command execution.

Task: 4403

### DIFF
--- a/cetmix_tower_git/i18n/hr.po
+++ b/cetmix_tower_git/i18n/hr.po
@@ -6,7 +6,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: Odoo Server 14.0\n"
 "Report-Msgid-Bugs-To: \n"
-"PO-Revision-Date: 2025-03-03 12:07+0000\n"
+"PO-Revision-Date: 2025-03-06 09:11+0000\n"
 "Last-Translator: Bole <bole@dajmi5.com>\n"
 "Language-Team: Croatian <https://hosted.weblate.org/projects/"
 "tower-server-14-0-dev/cetmix_tower_git/hr/>\n"
@@ -27,6 +27,10 @@ msgid ""
 "# %(vars)s \n"
 "# and run git-aggregator with '--expand-env' parameter.\n"
 msgstr ""
+"\n"
+"# Potrebno je postaviti sljedeće varijable u vaše okruženje:\n"
+"# %(vars)s \n"
+"# i pokrenuti git-aggregator sa ' --expand-env' parametrom\n"
 
 #. module: cetmix_tower_git
 #: code:addons/cetmix_tower_git/models/cx_tower_git_project.py:0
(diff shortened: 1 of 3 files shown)


cetmix_tower_server/wizards/cx_tower_command_execute_wizard.py
```diff
@@ -191,7 +202,7 @@ def execute_command_on_server(self):
         for server in self.server_ids:
             server.execute_command(
                 self.command_id,
-                sudo=self.use_sudo,
+                sudo=server.use_sudo,
```
ivs-cetmix: Why do we need "use_sudo" field in the wizard if it's not used anyway?

cetmix_tower_server/wizards/cx_tower_command_execute_wizard.py
```diff
@@ -272,7 +284,7 @@ def execute_command_in_wizard(self):
                     server._get_ssh_client(raise_on_error=True),
                     self.rendered_code,
                     self.path or None,
-                    sudo=self.use_sudo if self.use_sudo else None,
+                    sudo=server.use_sudo if server.use_sudo else None,
```
ivs-cetmix: Same here

ivs-cetmix (changes requested):

cetmix_tower_server/wizards/cx_tower_command_execute_wizard.py
```diff
@@ -179,7 +179,13 @@ def _check_sudo_mode_match(self):
         """
         Checks that the wizard's sudo mode matches the server configuration.
         """
-        for server in self.server_ids:
+        # If the user is in the 'group_user', use sudo rights to read the 'use_sudo'
+        # field since it's normally restricted to 'group_manager'.
+        if self.env.user.has_group("cetmix_tower_server.group_user"):
+            servers = self.server_ids.sudo()
```
ivs-cetmix: Why should we remove all access checks by using `sudo` here?

ivs-cetmix (changes requested):

cetmix_tower_server/wizards/cx_tower_command_execute_wizard.py
halbtonjazz: thanks for the idea, we've fixed it in the new commit

cetmix_tower_server/wizards/cx_tower_command_execute_wizard.py
halbtonjazz: thanks for the idea, we've fixed it in the new commit

cetmix_tower_server/wizards/cx_tower_command_execute_wizard.py
halbtonjazz: we discussed that it is better not to do that and moved the constrain from the wizard model to a method that runs commands

PR #257, opened 2025-04-17
[14.0][ADD] cetmix_tower_server-partner_interface

Mohajiro:
Add new partner interface to easily manage Cetmix Tower related entities:

- New 'Partners' submenu in 'Servers' menu
- 'Cetmix Tower' tab in partner form showing all secret keys of this partner.
- 'Servers' button in partner form with that shows the number of servers of this partner.

Task: 4548

## Summary by CodeRabbit

- **New Features**
  - Added a "Partners" menu item under the "Servers" menu for easier navigation and management of partner records.
  - Added a "Servers" stat button to the partner form, showing the count of related servers and enabling quick access to server records.
  - Introduced a new "Cetmix Tower" tab in the partner form to manage secret key values, with secret values masked by default.

- **Bug Fixes**
  - Ensured correct server count aggregation for partners, including those with servers assigned to child or descendant partners.
  - Fixed visibility of the "Servers" button to display only when relevant servers exist.

- **Tests**
  - Added comprehensive tests to verify server counting, including partners with no direct servers but servers on child partners.


cetmix_tower_server/views/cx_tower_partner_view.xml
```diff
@@ -0,0 +1,42 @@
...
+              </tree>
+              <form>
+                <group class="o_full_width">
+                  <field name="key_id" />
+                  <field name="secret_value" password="True" />
+                </group>
+              </form>
+            </field>
+          </page>
+        </xpath>
```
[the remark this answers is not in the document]
Mohajiro: Can you check cetmix_tower_server/views/cx_tower_partner_view.xml from last commit, all changes have been done

cetmix_tower_server/views/menuitems.xml
```diff
@@ -31,6 +31,15 @@
...
         sequence="10"
     />
+    <menuitem
+        id="menu_cx_tower_partner"
+        name="Partners"
+        action="action_cx_cetmix_tower_partner"
+        parent="menu_cx_tower_server_root"
+        groups="cetmix_tower_server.group_user"
+        sequence="15"
+    />
```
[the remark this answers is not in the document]
Mohajiro: I keep the full module path (cetmix_tower_server.group_user) rather than the shorthand group_user to avoid any ambiguity or accidental collisions with similarly named groups from other addons. By using the fully qualified reference, we guarantee that Odoo will always resolve to our own group record in the cetmix_tower_server namespace, even if another module elsewhere happens to define a group_user.

cetmix_tower_server/data/ir_actions_server.xml
```diff
@@ -23,4 +23,12 @@
         <field name="groups_id" eval="[(4, ref('cetmix_tower_server.group_user'))]" />
     </record>
 
+    <record id="action_cx_cetmix_tower_partner" model="ir.actions.act_window">
+      <field name="name">Partners</field>
+      <field name="type">ir.actions.act_window</field>
+      <field name="res_model">res.partner</field>
+      <field name="view_mode">tree,form</field>
+      <field name="groups_id" eval="[(4, ref('cetmix_tower_server.group_user'))]" />
+    </record>
```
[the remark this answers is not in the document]
Mohajiro: I see where you're coming from, but we actually designed it this way intentionally. The partners list needs to show all contacts because:

-  Admins often need to assign new servers to partners who don't have any yet
-     The server count field already makes it easy to see who has servers
-     We want to keep it consistent with how partners work elsewhere in Odoo

The security is already handled by the group restrictions, and partners' sensitive data like secrets have additional protections.

cetmix_tower_server/models/cx_tower_partner.py
```diff
@@ -0,0 +1,40 @@
...
+
+    @api.depends("server_ids", "child_ids")
+    def _compute_server_count(self):
+        """Compute the total number of servers for a partner,
+        including all nested child partners, using child_of."""
+        Server = self.env["cx.tower.server"]
+        for partner in self:
+            partner.server_count = Server.search_count(
+                [("partner_id", "child_of", partner.id)]
+            )
```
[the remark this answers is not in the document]
Mohajiro: We’re now using a stored field with the child_of operator in the domain, so recalculation happens correctly on both direct and descendant server changes. Adding an explicit child_ids.server_ids dependency is redundant here.

cetmix_tower_server/tests/test_partner_server_btn.py
```diff
@@ -0,0 +1,125 @@
...
+        """
+        ctx = self.partner_b.with_context(show_secret_value=True).secret_ids
+        self.assertEqual(len(ctx), 1)
+        self.assertEqual(ctx[0].secret_value, "TOPSECRET")
+
+        plain = self.partner_b.secret_ids
+        self.assertEqual(len(plain), 1)
+        self.assertEqual(
+            plain[0].secret_value, "TOPSECRET", "secret_value should always be exposed"
+        )
```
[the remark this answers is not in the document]
Mohajiro: I appreciate the suggestion to explicitly test show_secret_value=False, but in our implementation the placeholder substitution only happens during read() calls, not on direct field access. Adding a test here would require using read(['secret_value']) instead of accessing record.secret_value, which complicates the test for little gain. Given our current coverage—verifying that show_secret_value=True exposes the secret—this negative case doesn’t add practical value.

cetmix_tower_server/tests/test_partner_server_btn.py
```diff
@@ -0,0 +1,127 @@
...
+        self.Server.create({"partner_id": gc.id, **self.server_defaults})
+        self.Server.create({"partner_id": ggc.id, **self.server_defaults})
+
+        self.assertEqual(self.partner_b.server_count, 5)
+        self.assertEqual(gc.server_count, 2)
+        self.assertEqual(ggc.server_count, 1)
+
+    def test_parent_with_only_child_servers(self):
+        """Parent witthout servers directs and with child_of."""
```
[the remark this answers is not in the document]
Mohajiro: I see, thx U, I will fix it later

cetmix_tower_server/tests/test_partner_server_btn.py
```diff
@@ -0,0 +1,141 @@
...
+            masked[0].secret_value,
+            placeholder,
+            "Without show_secret_value context, secret_value must be masked",
+        )
+
+        # With explicit flag: real value
+        real = self.partner_b.with_context(show_secret_value=True).secret_ids
+        self.assertEqual(len(real), 1)
+        self.assertEqual(real[0].secret_value, "TOPSECRET")
+
```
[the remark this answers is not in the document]
Mohajiro: SECRET_VALUE_PLACEHOLDER is defined on the cx.tower.key model, not on cx.tower.key.value. Switching it to key.value would break the test, so we’ll keep the existing reference to cx.tower.key.

cetmix_tower_server/tests/test_partner_server_btn.py
```diff
@@ -0,0 +1,127 @@
...
+                "secret_value": "TOPSECRET",
+            }
+        )
+
+    def test_server_count_compute(self):
+        """Server count: direct + one‑level child + zero if none."""
+        self.assertEqual(self.partner_b.server_count, 3)
+        self.assertEqual(self.partner_b_child.server_count, 1)
+        self.assertEqual(self.partner_a.server_count, 0)
+
```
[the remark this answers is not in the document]
Mohajiro: The existing docstring already clearly states the three cases under test, and expanding it into a bullet list would only add noise without improving understanding. We'll keep the simpler form.

cetmix_tower_server/tests/test_partner_server_btn.py
```diff
@@ -0,0 +1,127 @@
...
+        plain = self.partner_b.secret_ids
+        self.assertEqual(len(plain), 1)
+        self.assertEqual(plain[0].secret_value, "TOPSECRET")
+
+    def test_button_visibility(self):
+        """
+        The 'Servers' stat button must always exist, be invisible when
+        server_count == 0, and visible when > 0.
+        """
```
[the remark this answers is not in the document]
Mohajiro: The current docstring is already clear and concise about the visibility conditions being tested, and expanding it into a bulleted list would add verbosity without much gain in readability. We'll keep the simpler form.

cetmix_tower_server/data/ir_actions_server.xml
```diff
@@ -23,4 +23,12 @@
         <field name="groups_id" eval="[(4, ref('cetmix_tower_server.group_user'))]" />
     </record>
 
+    <record id="action_cx_cetmix_tower_partner" model="ir.actions.act_window">
+      <field name="name">Partners</field>
+      <field name="type">ir.actions.act_window</field>
+      <field name="res_model">res.partner</field>
+      <field name="view_mode">tree,form</field>
+      <field name="groups_id" eval="[(4, ref('cetmix_tower_server.group_user'))]" />
```
ivs-cetmix: We can omit this because "res.partner" is a generic model and there is no need to restrict access to it.
But this is not an issue.

cetmix_tower_server/models/cx_tower_partner.py
ivs-cetmix, line 1: rename the file into "res_partner" in accordance with the OCA Coding Guidelines

cetmix_tower_server/models/cx_tower_partner.py
```diff
@@ -0,0 +1,40 @@
...
+        "partner_id",
+        string="Secrets",
+    )
+
+    @api.depends("server_ids", "child_ids")
+    def _compute_server_count(self):
+        """Compute the total number of servers for a partner,
+        including all nested child partners, using child_of."""
+        Server = self.env["cx.tower.server"]
+        for partner in self:
```
ivs-cetmix: You already have all servers of the current record in the `server_ids`. So you can simply do [like this](https://github.com/cetmix/cetmix-tower/blob/964aa82fdb34f9e030a01db1fc22e183f84744e4/cetmix_tower_server/models/cx_tower_variable.py#L132).

If you wan to compute all the servers including the child records, you can simply sum the `server_count` values of all child records.

```python
for partner in self:
    server_count = len(partner.server_ids)
    if partner.child_ids:
        server_count += sum(child.server_count for child in partner.child_ids)
```

This is approach is better because it may benefit from using the cached records instead of doing a new query for each record every time.

cetmix_tower_server/models/cx_tower_partner.py
```diff
@@ -0,0 +1,40 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+from odoo import api, fields, models
+
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    server_ids = fields.One2many("cx.tower.server", "partner_id", string="Servers")
+    server_count = fields.Integer(
+        "Number of Servers", compute="_compute_server_count", store=True
```
ivs-cetmix: 1. Odoo will generate field label from the field name "server_count" -> "Server Count". So you may skip adding a field label explicitly.
2. Using stored field means that it will be computed using `sudo` (until Odoo 18.0). Which means that it will show the number of all servers, even if user doesn't have access to them.
This can be misleading, because you see eg `5 Servers` and when you click the button there is only 2 servers available.
So there is no practical need of making this field stored.

cetmix_tower_server/models/cx_tower_partner.py
```diff
@@ -0,0 +1,40 @@
+# Copyright (C) 2025 Cetmix OÜ
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+from odoo import api, fields, models
+
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    server_ids = fields.One2many("cx.tower.server", "partner_id", string="Servers")
```
ivs-cetmix: This field is related to the model that has user restriction. So it's a good practice to apply a similar restriction to the fields as well.
`groups="cetmix_tower_server.group_user"`
Same for all other fields.

cetmix_tower_server/views/cx_tower_partner_view.xml
```diff
@@ -0,0 +1,42 @@
...
+      <field name="name">res.partner.form.inherit.cetmix.tower</field>
+      <field name="model">res.partner</field>
+      <field name="inherit_id" ref="base.view_partner_form" />
+      <field name="priority" eval="20" />
+
+      <field name="arch" type="xml">
+        <xpath expr="//div[@name='button_box']" position="inside">
+          <button
+                        name="action_view_partner_servers"
+                        type="object"
```
ivs-cetmix, line 13: Limit the button visibility to specific group members.

cetmix_tower_server/views/cx_tower_partner_view.xml
```diff
@@ -0,0 +1,42 @@
...
+                        class="oe_stat_button"
+                        icon="fa-server"
+                        attrs="{'invisible': [('server_count', '=', 0)]}"
+                    >
+            <field name="server_count" widget="statinfo" string="Servers" />
+          </button>
+        </xpath>
+
+        <xpath expr="//sheet/notebook" position="inside">
+          <page string="Cetmix Tower">
```
ivs-cetmix: Limit the page visibility to specific group members.

ivs-cetmix (changes requested):

cetmix_tower_server/data/ir_actions_server.xml
Mohajiro: Thanks you for information!

cetmix_tower_server/models/cx_tower_partner.py
Mohajiro: Done

cetmix_tower_server/models/cx_tower_partner.py
Mohajiro: Good idea, done.

cetmix_tower_server/models/res_partner.py
```diff
@@ -0,0 +1,49 @@
...
+    )
+
+    @api.depends("server_ids", "child_ids")
+    def _compute_server_count(self):
+        """Compute the total number of servers for a partner,
+        including all nested child partners, using child_of."""
+        for partner in self:
+            count = len(partner.server_ids)
+            count += sum(child.server_count for child in partner.child_ids)
+            partner.server_count = count
```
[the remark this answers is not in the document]
Mohajiro: I decided to use this version
@api.depends("server_ids", "child_ids")
def _compute_server_count(self):
    for partner in self:
        all_partners = self.env["res.partner"].search([("id", "child_of", partner.id)])
        partner.server_count = self.env["cx.tower.server"].search_count([
            ("partner_id", "in", all_partners.ids)
        ])
 because it's simpler and easier to read, iIt avoids unnecessary search queries inside the loop — child_ids are already available on the record, it's already covered by tests and works correctly, even for deep hierarchies.

cetmix_tower_server/views/cx_tower_partner_view.xml
Mohajiro: Ready

cetmix_tower_server/views/cx_tower_partner_view.xml
Mohajiro: Done

cetmix_tower_server/models/cx_tower_partner.py
Mohajiro: Done for all fields except server_count, because adding group restrictions on it broke view generation with
_Field server_count used in attrs.invisible [...] must be present in view but is missing._

cetmix_tower_server/models/res_partner.py
```diff
@@ -0,0 +1,49 @@
...
+        string="Servers",
+        groups="cetmix_tower_server.group_user",
+    )
+
+    server_count = fields.Integer(
+        "Number of Servers",
+        compute="_compute_server_count",
+        store=True,
+        # groups="cetmix_tower_server.group_user",
+    )
```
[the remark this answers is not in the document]
Mohajiro: It's make some error with view generation 
**Field server_count used in attrs.invisible [...] must be present in view but is missing.**

cetmix_tower_server/views/cx_tower_partner_view.xml
```diff
@@ -0,0 +1,62 @@
...
+                        attrs="{'invisible': [('server_count', '=', 0)]}"
+                    >
+            <field name="server_count" widget="statinfo" string="Servers" />
+          </button>
+        </xpath>
+
+        <xpath expr="//sheet/notebook" position="inside">
+          <page string="Cetmix Tower" groups="cetmix_tower_server.group_user">
+            <group string="Secrets">
+              <field
```
Bearnard21: We add new page inside the notebook. Ok.

        <xpath expr="//sheet/notebook" position="inside">
          <page string="Cetmix Tower" groups="cetmix_tower_server.group_user">
            <group name="group_tower> <- We add a group. I will take 100% of page. Don't add names it's just for eg.
              <group name="group_secrets" string="Secrets"> <- But if we add another group inside it, it will split the parent group in half. 
                <field 
                        name="secret_ids" nolabel="1" 
                        class="o_full_width"  < - I'm not sure we need that class here
                        domain="[('key_id.key_type','=','s')]" 
                        context="{'show_secret_value': True}"> <- Here you recall IDS field and want to see the related "cx.tower.key.value" model fields in a tree view, that's why you don't puta  closing tag 
                  <tree editable="bottom" class="o_list_full_width"> <-again this class
                    <field name="key_id" string="Key"/>
                    <field name="server_id" string="Server"/>
                    <field name="secret_value" string="Secret Value"/>
                  </tree> <-close the tree when all the needed fields are there
                </field> <- close the field. We don't need any form view inside the field, you've made it editable "bottom", it means we edit all the data without recalling form view, so please remove it.
              </group>
            </group>
          </page>
        </xpath>
        
        
Screenshot 2025-04-20 at 22 00 12

Bearnard21 (changes requested):

cetmix_tower_server/views/cx_tower_partner_view.xml
Mohajiro: Thank you for explication!

cetmix_tower_server/views/cx_tower_partner_view.xml
```diff
@@ -0,0 +1,53 @@
...
+              <group name="group_secrets" string="Secrets">
+                  <field
+                                    name="secret_ids"
+                                    nolabel="1"
+                                    domain="[('key_id.key_type', '!=', 's')]"
+                                    context="{'show_secret_value': True}"
+                                >
+                      <tree editable="bottom">
+                          <field name="key_id" string="Key" />
+                          <field name="server_id" string="Server" />
```
ivs-cetmix: No need to add "string" here, it will be added automatically based on the field name or string in the Python code.

cetmix_tower_server/views/cx_tower_partner_view.xml
```diff
@@ -0,0 +1,53 @@
...
+            <group>
+              <group name="group_secrets" string="Secrets">
+                  <field
+                                    name="secret_ids"
+                                    nolabel="1"
+                                    domain="[('key_id.key_type', '!=', 's')]"
+                                    context="{'show_secret_value': True}"
+                                >
+                      <tree editable="bottom">
+                          <field name="key_id" string="Key" />
```
ivs-cetmix: No need to add "string" here, unless you want to have some custom one. It will be added automatically based on the field name or string in the Python code.

cetmix_tower_server/views/cx_tower_partner_view.xml
```diff
@@ -0,0 +1,53 @@
...
+                          <field name="server_id" string="Server" />
+                          <field name="secret_value" string="Secret Value" />
```
ivs-cetmix: No need to add "string" here, unless you want to have some custom one. It will be added automatically based on the field name or string in the Python code.

cetmix_tower_server/views/cx_tower_partner_view.xml
```diff
@@ -0,0 +1,53 @@
...
+        </xpath>
+
+        <xpath expr="//sheet/notebook" position="inside">
+          <page string="Cetmix Tower" groups="cetmix_tower_server.group_user">
+            <group>
+              <group name="group_secrets" string="Secrets">
+                  <field
+                                    name="secret_ids"
+                                    nolabel="1"
+                                    domain="[('key_id.key_type', '!=', 's')]"
```
ivs-cetmix: You should put domain in the field definition in the Python code whenever it's possible.

cetmix_tower_server/tests/test_partner_server_btn.py
```diff
@@ -0,0 +1,127 @@
...
+        ctx = self.partner_b.with_context(show_secret_value=True).secret_ids
+        self.assertEqual(len(ctx), 1)
+        self.assertEqual(ctx[0].secret_value, "TOPSECRET")
+
+        plain = self.partner_b.secret_ids
+        self.assertEqual(len(plain), 1)
+        self.assertEqual(plain[0].secret_value, "TOPSECRET")
+
+    def test_button_visibility(self):
```
ivs-cetmix: We are not modifying this method, so no need to test it.

cetmix_tower_server/tests/test_partner_server_btn.py
```diff
@@ -0,0 +1,127 @@
...
+
+    def test_action_view_partner_servers_domain(self):
+        """Action domain must use 'child_of' operator for full hierarchy."""
+        action = self.partner_b.action_view_partner_servers()
+        self.assertEqual(
+            action.get("domain"),
+            [("partner_id", "child_of", self.partner_b.id)],
+        )
+
+    def test_secret_ids_relation(self):
```
ivs-cetmix: We are not modifying this method, so no need to test it.

cetmix_tower_server/tests/test_partner_server_btn.py
```diff
@@ -0,0 +1,127 @@
...
+    def test_parent_with_only_child_servers(self):
+        """Parent witthout servers directs and with child_of."""
+        parent = self.env["res.partner"].create({"name": "Parent Only"})
+        child = self.env["res.partner"].create(
+            {"name": "Child with Server", "parent_id": parent.id}
+        )
+        self.Server.create({"partner_id": child.id, **self.server_defaults})
+        self.assertEqual(parent.server_count, 1)
+
+    def test_action_view_partner_servers_domain(self):
```
ivs-cetmix: This method has hight probability of being modified later in custom modules. This may lead to this test failed, so no need to add this test at all.

cetmix_tower_server/tests/test_partner_server_btn.py
```diff
@@ -0,0 +1,127 @@
...
+            }
+        )
+
+    def test_server_count_compute(self):
+        """Server count: direct + one‑level child + zero if none."""
+        self.assertEqual(self.partner_b.server_count, 3)
+        self.assertEqual(self.partner_b_child.server_count, 1)
+        self.assertEqual(self.partner_a.server_count, 0)
+
+    def test_deep_hierarchy_server_count(self):
```
ivs-cetmix: What is this test used for?

cetmix_tower_server/models/res_partner.py
```diff
@@ -0,0 +1,46 @@
...
+        "cx.tower.key.value",
+        "partner_id",
+        string="Secrets",
+        groups="cetmix_tower_server.group_user",
+    )
+
+    @api.depends("server_ids", "child_ids.server_count")
+    def _compute_server_count(self):
+        for partner in self:
+            own = len(partner.server_ids)
```
ivs-cetmix: You should use more readable and explicit names for variables. Eg `own_server_count` and `child_server_count`

cetmix_tower_server/models/res_partner.py
```diff
@@ -0,0 +1,46 @@
...
+        "cx.tower.server",
+        "partner_id",
+        string="Servers",
+        groups="cetmix_tower_server.group_user",
+    )
+
+    server_count = fields.Integer(
+        "Number of Servers",
+        compute="_compute_server_count",
+        store=True,
```
ivs-cetmix: Once again: do we need to store this value?  Stored valued are computed using `sudo` until Odoo 18.0
This may lead to the wrong number of servers shown to users who don't have access to all servers of the partner.

cetmix_tower_server/models/res_partner.py
```diff
@@ -0,0 +1,46 @@
...
+
+    server_ids = fields.One2many(
+        "cx.tower.server",
+        "partner_id",
+        string="Servers",
+        groups="cetmix_tower_server.group_user",
+    )
+
+    server_count = fields.Integer(
+        "Number of Servers",
```
ivs-cetmix: This field depends on the `server_ids` field which is limited to specific groups. So this field should have the same restriction.

ivs-cetmix (changes requested):

cetmix_tower_server/views/cx_tower_partner_view.xml
Mohajiro: Thanks you, it's done

cetmix_tower_server/views/cx_tower_partner_view.xml
Mohajiro: Done

cetmix_tower_server/views/cx_tower_partner_view.xml
Mohajiro: Done

cetmix_tower_server/views/cx_tower_partner_view.xml
Mohajiro: Thanks, I gonna keep it in mind

cetmix_tower_server/tests/test_partner_server_btn.py
Mohajiro: Thanks, it's done

cetmix_tower_server/tests/test_partner_server_btn.py
Mohajiro: Thank you, done

cetmix_tower_server/tests/test_partner_server_btn.py
Mohajiro: Ready

cetmix_tower_server/tests/test_partner_server_btn.py
Mohajiro: To verify the server count rolls up correctly through multiple hierarchy levels, I have removed this test.

cetmix_tower_server/models/res_partner.py
Mohajiro: Okey, thanks. Done

cetmix_tower_server/models/res_partner.py
Mohajiro: I’m having issues generating the view after restricting access to server_count. I tried giving the same access rights to the view elements, but it still doesn’t work.

ivs-cetmix: When you say "having issues", what do you mean exactly ?

Mohajiro: "Field server_count used in attrs.invisible ({'invisible': [('server_count', '=', 0)]}) must be present in view but is missing" 
This type of errors.

cetmix_tower_server/models/res_partner.py
```diff
@@ -0,0 +1,46 @@
...
+
+    def action_view_partner_servers(self):
+        """Open server list filtered by partner and all its descendants."""
+        self.ensure_one()
+        return {
+            "name": "Servers",
+            "type": "ir.actions.act_window",
+            "res_model": "cx.tower.server",
+            "view_mode": "kanban,tree,form",
+            "domain": [("partner_id", "child_of", self.id)],
```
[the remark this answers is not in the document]
Mohajiro: I see but I prefer de keep my version

cetmix_tower_server/models/res_partner.py
```diff
@@ -0,0 +1,46 @@
...
+        "cx.tower.server",
+        "partner_id",
+        string="Servers",
+        groups="cetmix_tower_server.group_user",
+    )
+
+    server_count = fields.Integer(
+        "Number of Servers",
+        compute="_compute_server_count",
+    )
```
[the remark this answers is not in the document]
Mohajiro: Okey, thanks.

cetmix_tower_server/tests/test_partner_server_btn.py
```diff
@@ -0,0 +1,66 @@
...
+
+@tagged("partner_servers_btn")
+class TestPartnerServers(TestTowerCommon):
+    """
+    Tests for the res.partner → cx.tower.server integration:
+      - server_count computation (direct, child-of, deep-hierarchy)
+      - action domain uses 'child_of'
+      - secret_ids unmasking via context
+      - 'Servers' stat button visibility attrs
+    """
```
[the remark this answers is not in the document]
Mohajiro: Resolved

cetmix_tower_server/models/res_partner.py
Mohajiro: Thanks you, deleted.

cetmix_tower_server/views/cx_tower_partner_view.xml
```diff
@@ -0,0 +1,52 @@
...
+        </xpath>
+
+        <xpath expr="//sheet/notebook" position="inside">
+          <page string="Cetmix Tower" groups="cetmix_tower_server.group_user">
+            <group>
+              <group name="group_secrets" string="Secrets">
+                  <field
+                                    name="secret_ids"
+                                    nolabel="1"
+                                    context="{'show_secret_value': True}"
```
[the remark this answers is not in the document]
Mohajiro: Done

cetmix_tower_server/models/res_partner.py
ivs-cetmix: Because you need to add the same `groups` on the view element this condition is used in.

cetmix_tower_server/views/cx_tower_partner_view.xml
```diff
@@ -33,13 +33,12 @@
                   <field
                                     name="secret_ids"
                                     nolabel="1"
-                                    domain="[('key_id.key_type', '!=', 's')]"
                                     context="{'show_secret_value': True}"
```
ivs-cetmix: What this context key is used for?

ivs-cetmix (changes requested):

cetmix_tower_server/views/cx_tower_partner_view.xml
Mohajiro: It's return just "secret" records with key_type='s' from cx.tower.key.value, but right now I put this line in the model.

cetmix_tower_server/models/res_partner.py
```diff
@@ -0,0 +1,45 @@
...
+    server_ids = fields.One2many(
+        "cx.tower.server",
+        "partner_id",
+        string="Servers",
+        groups="cetmix_tower_server.group_user",
+    )
+
+    server_count = fields.Integer(
+        compute="_compute_server_count",
+    )
```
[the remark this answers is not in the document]
Mohajiro: For now we can keep it like this

cetmix_tower_server/views/cx_tower_partner_view.xml
ivs-cetmix: This is not fixed yet

ivs-cetmix (changes requested):

cetmix_tower_server/views/cx_tower_partner_view.xml
Mohajiro: Fixed

Bearnard21 (approved):
Functionlity LGTM

cetmix_tower_server/views/cx_tower_partner_view.xml
```diff
@@ -0,0 +1,43 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+  <data>
+    <record id="view_res_partner_form_inherit_cetmix_tower" model="ir.ui.view">
+      <field name="name">res.partner.form.inherit.cetmix.tower</field>
+      <field name="model">res.partner</field>
+      <field name="inherit_id" ref="base.view_partner_form" />
+      <field name="priority" eval="20" />
```
ivs-cetmix: Note: do not put priority explicitly  in the view inheritance order, unless you understand why exactly you put this particular number.

ivs-cetmix (changes requested):

ivs-cetmix (commented):
Check the comment please, not sure if this is done for some reason.

cetmix_tower_server/views/cx_tower_partner_view.xml
Mohajiro: Done, thank you.

ivs-cetmix (approved):
/ocabot merge nobump

CetmixGitDrone:
Prepared branch [14.0-ocabot-merge-pr-257-by-ivs-cetmix-bump-nobump](https://github.com/cetmix/cetmix-tower/commits/14.0-ocabot-merge-pr-257-by-ivs-cetmix-bump-nobump), awaiting test results.

CetmixGitDrone:
Congratulations, your PR was merged at c960687227257180d3afba0be55292b1a77bb946. Thanks a lot for contributing to cetmix. ❤️

PR #309, opened 2025-06-03
[14.0][IMP] cetmix_tower_server: file create command update

Kurama2d:
Add `if_file_exists` and `disconnect_file` fields to "Create file using template" command. 
Supports skipping, overwriting, or raising an error when uploading a file that already exists.

Task: 4740

## Summary by CodeRabbit

- **New Features**
  - Added options to control file creation behavior when a file already exists on the server: skip, overwrite, or raise an error.
  - Introduced a setting to disconnect created files from their templates after command execution.
- **Bug Fixes**
  - Enhanced file-template linkage handling based on new command settings.
- **Tests**
  - Added tests covering file creation behaviors and template disconnection.
- **User Interface**
  - Updated command form to include new fields for file existence handling and template disconnection, shown for applicable actions.

### DIFF
--- a/cetmix_tower_server/models/cx_tower_command.py
+++ b/cetmix_tower_server/models/cx_tower_command.py
@@ -155,6 +155,26 @@ class CxTowerCommand(models.Model):
         column2="variable_id",
     )
 
+    if_file_exists = fields.Selection(
+        selection=[
+            ("skip", "Skip"),
+            ("overwrite", "Overwrite"),
+            ("raise", "Raise Error"),
+        ],
+        default="skip",
+        help="What to do if file already exists on the server.\n"
+        "- Skip: Do not create or update the file.\n"
+        "- Overwrite: Replace the existing file with the new one.\n"
+        "- Raise Error: Raise an error if the file already exists.",
+    )
+    disconnect_file = fields.Boolean(
+        string="Disconnect from Template",
+        help=(
+            "If enabled, disconnects the file from its template "
+            "after running the command.\n"
+        ),
+    )
+
     # ---- Access. Add relation for mixin fields
     user_ids = fields.Many2many(
         relation="cx_tower_command_user_rel",
--- a/cetmix_tower_server/models/cx_tower_file_template.py
+++ b/cetmix_tower_server/models/cx_tower_file_template.py
@@ -127,7 +127,7 @@ class CxTowerFileTemplate(models.Model):
         return action
 
     # -- Business logic
-    def create_file(self, server, server_dir="", raise_if_exists=False):
+    def create_file(self, server, server_dir="", if_file_exists="raise"):
         """
         Create a new file using the current template for the selected server.
         If the same file already exists, just ignore it or raise an error based on the
@@ -136,45 +136,86 @@ class CxTowerFileTemplate(models.Model):
         :param server: recordset
             The server (cx.tower.server) on which the file should be created. This is a
             required parameter.
+        :param if_file_exists: str, optional
+            Defines the behavior if the file already exists on the server.
         :param server_dir: str, optional
             The directory on the server where the file should be created. If not set,
             the server_dir field of the template will be used.
-        :param raise_if_exists: bool, optional
-            If set to True, raises a ValidationError if the file already exists on the
-            server. Defaults to False.
 
         :return: cx.tower.file
             Returns the newly created file record (cx.tower.file) if the file was
-            created successfully. Returns the existing file record if the file already
-            exists and raise_if_exists is False.
+            created successfully or if_file_exists is set to "overwrite".
+            Returns the existing file record if the file already exists
+            and if_file_exists is set to "skip".
 
         :raises ValidationError:
-            If the file already exists on the server and raise_if_exists is True.
+            If the file already exists on the server if_file_exists is set to "raise".
         """
         self.ensure_one()
+        # Explicit guard against invalid behavior values
+        valid_behaviors = {"skip", "raise", "overwrite"}
+        if if_file_exists not in valid_behaviors:
+            raise ValidationError(
+                f"Invalid if_file_exists value: {if_file_exists}. "
+                f"Expected one of {valid_behaviors}."
+            )
         file_model = self.env["cx.tower.file"]
-        existing_file = file_model.search(
+        existing_files = file_model.search(
             [
-                ("template_id", "=", self.id),
                 ("server_id", "=", server.id),
-                ("server_dir", "=", server_dir or self.server_dir),
                 ("source", "=", self.source),
             ],
-            limit=1,
+            order="id DESC",
         )
-        if existing_file:
-            if raise_if_exists:
-                raise ValidationError(_("File already exists on server."))
+        existing_dir = server_dir or self.server_dir
+
+        # Render the server directory and file name from the template
+        variables = list(
+            set(
+                self.get_variables_from_code(self.file_name)
+                + self.get_variables_from_code(existing_dir)
+            )
+        )
+        var_vals = server.get_variable_values(variables).get(server.id) or {}
+
+        unrendered_path = f"{existing_dir}/{self.file_name}"
+        rendered_path = self.render_code_custom(unrendered_path, **var_vals)
+
+        # Filter existing files by rendered path
+        existing_files = existing_files.filtered(
+            lambda f: f.full_server_path == rendered_path
+        )
+
+        # Filter existing files by template if it exists, otherwise take the first one
+        existing_file = (
+            existing_files.filtered(lambda f: f.template_id == self)[:1]
+            or existing_files[:1]
+        )
+
+        if existing_file and if_file_exists == "skip":
+            return existing_file.with_context(file_creation_skipped=True)
+
+        if existing_file and if_file_exists == "raise":
+            raise ValidationError(_("File already exists on server."))
+
+        if existing_file and if_file_exists == "overwrite":
+            existing_file.with_context(is_custom_server_dir=True).write(
+                {
+                    "template_id": self.id,
+                }
+            )
             return existing_file
 
-        return file_model.with_context(is_custom_server_dir=True).create(
-            {
-                "template_id": self.id,
-                "server_id": server.id,
-                "name": self.file_name,
-                "code_on_server": self.code,
-                "server_dir": server_dir or self.server_dir,
-                "file_type": self.file_type,
-                "source": self.source,
-            }
-        )
+        vals = {
+            "name": self.file_name,
+            "server_id": server.id,
+            "server_dir": existing_dir,
+            "template_id": self.id,
+            "code": self.code,
+            "file_type": self.file_type,
+            "source": self.source,
+        }
+
+        new_file = file_model.with_context(is_custom_server_dir=True).create(vals)
+        # Return new_file if no file exists
+        return new_file
--- a/cetmix_tower_server/models/cx_tower_server.py
+++ b/cetmix_tower_server/models/cx_tower_server.py
@@ -1113,7 +1113,6 @@ class CxTowerServer(models.Model):
         elif command.action == "file_using_template":
             response = self._command_runner_file_using_template(
                 log_record,
-                command.file_template_id,
                 rendered_command_path,
                 **kwargs,
             )
@@ -1165,7 +1164,6 @@ class CxTowerServer(models.Model):
     def _command_runner_file_using_template(
         self,
         log_record,
-        file_template_id,
         server_dir,
         **kwargs,
     ):
@@ -1181,8 +1179,6 @@ class CxTowerServer(models.Model):
         Args:
             log_record (recordset): The log record to update with the command's
                 status.
-            file_template_id (recordset): The file template to use for creating
-                the new file.
             server_dir (str): The directory on the server where the file should be
                 created.
             **kwargs: Additional keyword arguments.
@@ -1197,10 +1193,10 @@ class CxTowerServer(models.Model):
         """
         try:
             # Attempt to create a new file using the template for the current server
-            file = file_template_id.create_file(
+            file = log_record.command_id.file_template_id.create_file(
                 server=self,
+                if_file_exists=log_record.command_id.if_file_exists,
                 server_dir=server_dir,
-                raise_if_exists=True,
             )
 
             # If file creation failed, log the failure and exit
@@ -1220,16 +1216,32 @@ class CxTowerServer(models.Model):
                 else:
                     return command_result
 
-            if file.source == "server":
-                file.action_pull_from_server()
-            elif file.source == "tower":
-                file.action_push_to_server()
-            else:
-                raise UserError(
-                    _(
-                        "File source cannot be determined: '%(source)s'",
-                        source=file.source,
+            # Context is used to detect a retry
+            # and avoid handling skip logic on first attempt
+            is_creation_skipped = file._context.get("file_creation_skipped")
+
+            if not is_creation_skipped:
+                if file.source == "server":
+                    file.action_pull_from_server()
+                elif file.source == "tower":
+                    file.action_push_to_server()
+                else:
+                    raise UserError(
+                        _(
+                            "File source cannot be determined: '%(source)s'",
+                            source=file.source,
+                        )
                     )
+
+            if log_record.command_id.disconnect_file:
+                file.action_unlink_from_template()
+
+            if is_creation_skipped:
+                return log_record.finish(
+                    fields.Datetime.now(),
+                    0,
+                    _("File already exists on server. Upload skipped"),
+                    None,
                 )
 
             # Log the successful creation and upload of the file
--- a/cetmix_tower_server/tests/test_command.py
+++ b/cetmix_tower_server/tests/test_command.py
@@ -1265,6 +1265,110 @@ result = re.sub(pattern, replacement, value)
             command_result["error"], "Command error doesn't match expected one"
         )
 
+    def test_file_using_template_create_if_exists(self):
+        """Test uploading file using template if it exists on server."""
+
+        command = self.command_create_file_with_template_server_source
+        command.write({"if_file_exists": "skip"})
+
+        # Create file to make sure that it exists on the server
+        file_template = command.file_template_id
+        orig_file = file_template.create_file(
+            server=self.server_test_1,
+            server_dir=file_template.server_dir,
+            if_file_exists=command.if_file_exists,
+        )
+
+        self.assertTrue(orig_file, "File must be created on the server")
+
+        # Test if file exists and command is set to "skip"
+        skipped_file = file_template.create_file(
+            server=self.server_test_1,
+            server_dir=file_template.server_dir,
+            if_file_exists=command.if_file_exists,
+        )
+        self.assertEqual(
+            orig_file,
+            skipped_file,
+            "Skip should return the existing file, not create a new one",
+        )
+        self.assertEqual(
+            self.env["cx.tower.file"].search_count(
+                [
+                    ("template_id", "=", file_template.id),
+                    ("server_id", "=", self.server_test_1.id),
+                ]
+            ),
+            1,
+            "There must be exactly one physical file record after skip",
+        )
+
+        # Change command to raise an error if file exists
+        command.write({"if_file_exists": "raise"})
+        with self.assertRaisesRegex(
+            ValidationError,
+            "File already exists on server.",
+        ):
+            file_template.create_file(
+                server=self.server_test_1,
+                server_dir=file_template.server_dir,
+                if_file_exists=command.if_file_exists,
+            )
+        # Change command to "overwrite" file if it exists
+        command.write({"if_file_exists": "overwrite"})
+        # Run command again, it should overwrite the file
+        file_template.create_file(
+            server=self.server_test_1,
+            server_dir=file_template.server_dir,
+            if_file_exists=command.if_file_exists,
+        )
+        self.assertEqual(
+            self.env["cx.tower.file"].search_count(
+                [
+                    ("template_id", "=", file_template.id),
+                    ("server_id", "=", self.server_test_1.id),
+                    ("server_dir", "=", file_template.server_dir),
+                ]
+            ),
+            1,
+            "There must be exactly one physical file record after overwrite",
+        )
+        self.assertEqual(
+            orig_file.code,
+            file_template.code,
+            "File code must match template after overwrite",
+        )
+        self.assertEqual(
+            orig_file.name,
+            file_template.file_name,
+            "File name must match template after overwrite",
+        )
+        self.assertEqual(
+            orig_file.source,
+            file_template.source,
+            "File source must match template after overwrite",
+        )
+
+    def test_is_file_disconnected_from_template(self):
+        """Test if file is disconnected from template after being created."""
+
+        initial_files = self.server_test_1.file_ids
+        command = self.command_create_file_with_template_server_source
+
+        command.disconnect_file = True
+        self.server_test_1.run_command(command=command)
+
+        new_files = self.server_test_1.file_ids - initial_files
+        self.assertEqual(len(new_files), 1, "Must be one new file created")
+        self.assertEqual(
+            new_files.code_on_server,
+            command.file_template_id.code,
+            "File code must match template",
+        )
+        self.assertFalse(
+            new_files.template_id, "File must be disconnected from template"
+        )
+
     # ---------------------
     # *********************
     #   Python commands
--- a/cetmix_tower_server/tests/test_file.py
+++ b/cetmix_tower_server/tests/test_file.py
@@ -302,7 +302,11 @@ class TestTowerFile(TestTowerCommon):
             }
         )
 
-        file = file_template.create_file(server=self.server_test_1)
+        file = file_template.create_file(
+            server=self.server_test_1,
+            server_dir=file_template.server_dir,
+            if_file_exists="overwrite",
+        )
         self.assertEqual(file.code, self.file_template.code)
         self.assertEqual(file.template_id, file_template)
         self.assertEqual(file.server_id, self.server_test_1)
@@ -310,10 +314,16 @@ class TestTowerFile(TestTowerCommon):
         self.assertEqual(file.server_dir, self.file_template.server_dir)
 
         with self.assertRaises(exceptions.ValidationError):
-            file_template.create_file(server=self.server_test_1, raise_if_exists=True)
+            file_template.create_file(
+                server=self.server_test_1,
+                server_dir=file_template.server_dir,
+                if_file_exists="raise",
+            )
 
         another_file = file_template.create_file(
-            server=self.server_test_1, raise_if_exists=False
+            server=self.server_test_1,
+            server_dir=file_template.server_dir,
+            if_file_exists="skip",
         )
         self.assertEqual(another_file, file)
 
@@ -343,13 +353,13 @@ class TestTowerFile(TestTowerCommon):
             file_template.create_file(
                 server=self.server_test_1,
                 server_dir="/var/tmp/custom",
-                raise_if_exists=True,
+                if_file_exists="raise",
             )
 
         another_file = file_template.create_file(
             server=self.server_test_1,
             server_dir="/var/tmp/custom",
-            raise_if_exists=False,
+            if_file_exists="skip",
         )
         self.assertEqual(another_file, file)
 
--- a/cetmix_tower_yaml/tests/test_command.py
+++ b/cetmix_tower_yaml/tests/test_command.py
@@ -34,6 +34,8 @@ server_status: false
 variable_ids: false
 secret_ids: false
 no_split_for_sudo: false
+if_file_exists: skip
+disconnect_file: false
 """
 
         # YAML content translated into Python dict
@@ -111,6 +113,16 @@ Ensure all fields are rendered properly.""",
             self.command_test_yaml_dict["reference"],
             "YAML value doesn't match Cetmix Tower one",
         )
+        self.assertEqual(
+            command_test.if_file_exists,
+            self.command_test_yaml_dict["if_file_exists"],
+            "YAML value doesn't match Cetmix Tower one",
+        )
+        self.assertEqual(
+            command_test.disconnect_file,
+            self.command_test_yaml_dict["disconnect_file"],
+            "YAML value doesn't match Cetmix Tower one",
+        )
 
     def test_command_from_yaml(self):
         """Test if YAML is generated properly from a command"""
@@ -161,6 +173,16 @@ Ensure all fields are rendered properly.""",
             self.assertEqual(
                 command.reference,
                 self.command_test_yaml_dict["reference"],
+                "YAML value doesn't match Cetmix Tower one",
+            )
+            self.assertEqual(
+                command.if_file_exists,
+                self.command_test_yaml_dict["if_file_exists"],
+                "YAML value doesn't match Cetmix Tower one",
+            )
+            self.assertEqual(
+                command.disconnect_file,
+                self.command_test_yaml_dict["disconnect_file"],
                 "YAML value doesn't match Cetmix Tower one",
             )
 
@@ -248,6 +270,8 @@ server_status: false
 variable_ids: false
 secret_ids: false
 no_split_for_sudo: false
+if_file_exists: skip
+disconnect_file: false
 """
         # Add file template
         file_template = self.env["cx.tower.file.template"].create(
@@ -311,6 +335,8 @@ server_status: false
 variable_ids: false
 secret_ids: false
 no_split_for_sudo: false
+if_file_exists: skip
+disconnect_file: false
 """
         command_with_template.invalidate_cache(["yaml_code"])
         self.assertEqual(
--- a/cetmix_tower_server/models/cx_tower_server_template.py
+++ b/cetmix_tower_server/models/cx_tower_server_template.py
@@ -304,7 +304,7 @@ class CxTowerServerTemplate(models.Model):
         logs = server.server_log_ids.filtered(lambda rec: rec.log_type == "file")
         for log in logs.sudo():
             log.file_id = log.file_template_id.create_file(
-                server=server, raise_if_exists=False
+                server=server, if_file_exists="skip"
             ).id
 
         flight_plan = server.server_template_id.flight_plan_id
--- a/cetmix_tower_server/tests/common.py
+++ b/cetmix_tower_server/tests/common.py
@@ -163,6 +163,7 @@ class TestTowerCommon(TransactionCase):
                 "path": "/home/{{ tower.server.username }}",
                 "action": "file_using_template",
                 "file_template_id": self.template_file_tower.id,
+                "if_file_exists": "raise",
             }
         )
 
@@ -172,6 +173,7 @@ class TestTowerCommon(TransactionCase):
                 "path": "/home/{{ tower.server.username }}",
                 "action": "file_using_template",
                 "file_template_id": self.template_file_server.id,
+                "if_file_exists": "raise",
             }
         )
 
--- a/cetmix_tower_server/views/cx_tower_command_view.xml
+++ b/cetmix_tower_server/views/cx_tower_command_view.xml
@@ -71,6 +71,17 @@
                                 attrs="{'invisible': [('action', '=', 'python_code')]}"
                                 placeholder="optional, eg /home/{{ tower.server.username }}"
                             />
+                            <field
+                                name="if_file_exists"
+                                attrs="{
+                                    'invisible': [('action', '!=', 'file_using_template')],
+                                    'required': [('action', '=', 'file_using_template')],
+                                }"
+                            />
+                            <field
+                                name="disconnect_file"
+                                attrs="{'invisible': [('action', '!=', 'file_using_template')]}"
+                            />
                             <field name="allow_parallel_run" />
                             <field name="no_split_for_sudo" />
                             <field name="note" />
--- a/cetmix_tower_yaml/models/cx_tower_command.py
+++ b/cetmix_tower_yaml/models/cx_tower_command.py
@@ -25,5 +25,7 @@ class CxTowerCommand(models.Model):
             "variable_ids",
             "secret_ids",
             "no_split_for_sudo",
+            "if_file_exists",
+            "disconnect_file",
         ]
         return res
--- a/cetmix_tower_yaml/tests/test_yaml_export_wizard.py
+++ b/cetmix_tower_yaml/tests/test_yaml_export_wizard.py
@@ -122,6 +122,8 @@ records:
   code: echo 'Test Command From Yaml'
   server_status: false
   no_split_for_sudo: false
+  if_file_exists: skip
+  disconnect_file: false
 - cetmix_tower_model: command
   access_level: manager
   reference: test_command_from_yaml_2
@@ -133,6 +135,8 @@ records:
   code: echo 'Test Command From Yaml 2'
   server_status: false
   no_split_for_sudo: false
+  if_file_exists: skip
+  disconnect_file: false
 """
 
         # -- 1 --


cetmix_tower_server/models/cx_tower_file_template.py
```diff
@@ -127,7 +127,9 @@ def action_open_files(self):
         return action
 
     # -- Business logic
-    def create_file(self, server, server_dir="", raise_if_exists=False):
+    def create_file(
+        self, server, if_file_exists="raise", server_dir="", raise_if_exists=False
+    ):
```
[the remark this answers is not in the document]
Kurama2d: I prefer to keep order that had been here before my changes and only wanted to add new argument.

cetmix_tower_server/models/cx_tower_server.py
```diff
@@ -1199,6 +1201,7 @@ def _command_runner_file_using_template(
             # Attempt to create a new file using the template for the current server
             file = file_template_id.create_file(
                 server=self,
+                if_file_exists=command.if_file_exists,
                 server_dir=server_dir,
                 raise_if_exists=True,
             )
```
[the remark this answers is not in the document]
Kurama2d: We need raise_if_exists=True for the case when an error raises when if_file_exists is "raise" and raise_if_exists=True.

cetmix_tower_server/models/cx_tower_file_template.py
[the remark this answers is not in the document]
Kurama2d: I removed raise_if_exists param

cetmix_tower_server/models/cx_tower_server.py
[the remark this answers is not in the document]
Kurama2d: You was right, raise_if_exists is not needed anymore

cetmix_tower_server/models/cx_tower_file_template.py
```diff
@@ -127,7 +127,7 @@ def action_open_files(self):
         return action
 
     # -- Business logic
-    def create_file(self, server, server_dir="", raise_if_exists=False):
+    def create_file(self, server, if_file_exists="raise", server_dir=""):
```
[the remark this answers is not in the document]
Kurama2d: It's not really needed here.

cetmix_tower_server/models/cx_tower_command.py
```diff
@@ -140,6 +140,21 @@ class CxTowerCommand(models.Model):
...
+            ("raise", "Raise Error"),
+        ],
+        string="If File Exists",
+        default="skip",
+        help="What to do if file already exists on the server.\n"
+        "- Skip: Do not create or update the file.\n"
+        "- Overwrite: Replace existing file with new one.\n"
+        "- Raise Error: Raise an error if the file already exists.",
+    )
+    is_disconnect_from_tmpl = fields.Boolean(string="Disconnect from Template")
```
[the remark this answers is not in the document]
Kurama2d: Thanks, help is added

cetmix_tower_server/models/cx_tower_file_template.py
```diff
@@ -178,3 +179,5 @@ def create_file(self, server, server_dir="", raise_if_exists=False):
                 "source": self.source,
             }
         )
+        if not existing_file or (existing_file and if_file_exists == "overwrite"):
+            return new_file
```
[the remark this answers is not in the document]
Kurama2d: Thanks, made this change

cetmix_tower_server/tests/test_command.py
```diff
@@ -1265,6 +1265,58 @@ def test_run_command_no_command_log(self):
...
+            "File must be overwritten on the server",
+        )
+
+    def test_is_file_disconnected_from_template(self):
+        """Test if file is disconnected from template after being created."""
+        command = self.command_create_file_with_template_server_source
+        command.is_disconnect_from_tmpl = True
+        self.server_test_1.run_command(command=command)
+        template_id = self.server_test_1.file_ids.template_id
+        self.assertFalse(template_id, "File must be disconnected from template")
```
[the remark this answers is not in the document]
Kurama2d: Thanks, changed this logic

cetmix_tower_server/tests/test_command.py
```diff
@@ -1265,6 +1265,58 @@ def test_run_command_no_command_log(self):
...
+        file = file_template.create_file(
+            server=self.server_test_1,
+            if_file_exists=command.if_file_exists,
+        )
+        self.assertNotEqual(
+            orig_file,
+            file,
+            "File must be overwritten on the server",
+        )
+
```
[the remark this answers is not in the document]
Kurama2d: We need to write skip first to create new file, because we have default value 'raise' for this command and if there are any file already on the server to start with this case or if there is no file on the server just create new one and check 'skip' case

cetmix_tower_server/models/cx_tower_command.py
```diff
@@ -140,6 +140,24 @@ class CxTowerCommand(models.Model):
...
+        selection=[
+            ("skip", "Skip"),
+            ("overwrite", "Overwrite"),
+            ("raise", "Raise Error"),
+        ],
+        string="If File Exists",
+        default="skip",
+        help="What to do if file already exists on the server.\n"
+        "- Skip: Do not create or update the file.\n"
+        "- Overwrite: Replace existing file with new one.\n"
```
ivs-cetmix: "- Overwrite: Replace the existing the file with the new one.\n"

cetmix_tower_server/models/cx_tower_command.py
```diff
@@ -140,6 +140,24 @@ class CxTowerCommand(models.Model):
...
+        "- Raise Error: Raise an error if the file already exists.",
+    )
+    is_disconnect_from_tmpl = fields.Boolean(
+        string="Disconnect from Template",
+        help="If enabled, disconnect from the template after running the command.\n",
```
ivs-cetmix: help="If enabled, disconnect from the file template after running the command.\n",

cetmix_tower_server/models/cx_tower_file_template.py
ivs-cetmix: Restrain from switching function arguments position as it may lead to an unpredictable behaviour in the modules that inherit this one.

cetmix_tower_server/models/cx_tower_server.py
```diff
@@ -1232,6 +1234,9 @@ def _command_runner_file_using_template(
                     )
                 )
 
+            if command.is_disconnect_from_tmpl:
+                file.template_id = False
```
ivs-cetmix: Use [the existing function](https://github.com/cetmix/cetmix-tower/blob/ae60ad81c1da9554ecbfb849a2717ef3191c1009/cetmix_tower_server/models/cx_tower_file.py#L314) for that to preserve any custom logic it may implement.

ivs-cetmix (changes requested):

cetmix_tower_server/models/cx_tower_server.py
Kurama2d: Thanks, replaced to this func

cetmix_tower_server/models/cx_tower_file_template.py
Kurama2d: Thanks, moved if_file_exists after server_dir to avoid confusion.

cetmix_tower_server/models/cx_tower_command.py
Kurama2d: Thanks, changed a little

cetmix_tower_server/models/cx_tower_command.py
Kurama2d: Thanks, added articles to their positions

cetmix_tower_server/models/cx_tower_command.py
```diff
@@ -140,6 +140,27 @@ class CxTowerCommand(models.Model):
         column2="variable_id",
     )
 
+    if_file_exists = fields.Selection(
+        selection=[
+            ("skip", "Skip"),
+            ("overwrite", "Overwrite"),
+            ("raise", "Raise Error"),
+        ],
+        string="If File Exists",
```
ivs-cetmix: No need to add a string here explicitly, it will be [generated automatically](https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#fields) from the field name.

cetmix_tower_server/models/cx_tower_server.py
```diff
@@ -1165,6 +1166,7 @@ def _command_runner(
     def _command_runner_file_using_template(
         self,
         log_record,
+        command,
```
ivs-cetmix: Do we actually need the command? We can always get if from the `plan_log` record. 
Meanwhile we can drop `file_template` argument either  as we can also get if from the command later in the runner function.

ivs-cetmix (changes requested):

cetmix_tower_server/models/cx_tower_command.py
Kurama2d: Yes, you are right. Removed the string.

cetmix_tower_server/models/cx_tower_server.py
Kurama2d: Double checked, we don't really need it. Also removed file_template argument and replace it to call template from log and command directly.

cetmix_tower_server/tests/test_command.py
```diff
@@ -1265,6 +1265,76 @@ def test_run_command_no_command_log(self):
...
+    def test_is_file_disconnected_from_template(self):
+        """Test if file is disconnected from template after being created."""
+
+        initial_files = self.server_test_1.file_ids
+        command = self.command_create_file_with_template_server_source
+
+        command.is_disconnect_from_tmpl = True
+        self.server_test_1.run_command(command=command)
+
+        new_files = self.server_test_1.file_ids - initial_files
```
ivs-cetmix, line 1361: We need to ensure that the `new_files` is exactly the file that was uploaded with the command. Just to avoid any potential issues.

ivs-cetmix (changes requested):
Some minor comments, otherwise looks good.

GabbasovDinar (approved):
LGTM

cetmix_tower_server/tests/test_command.py
Kurama2d: Here in this case the only way to check that new_files is exactly the file that was uploaded with the command if the assert with empty template. We don't have template as the source of date to check connection with command after command run. We don't have any flag to ensure in command log, command or server.

ivs-cetmix: Agree, however we can check:
a) that there is only one file was created `len(new_files) == 1`
b) that that file has the same code as the template `file.code == template.code`

ivs-cetmix (changes requested):
Please add this new field into the list of exported fields in the [cetmix_tower_yaml](https://github.com/cetmix/cetmix-tower/blob/f5787bee25540a8f9c9269863db1d352a0a45d8d/cetmix_tower_yaml/models/cx_tower_command.py#L10) module. This should be done in  a separate commit.

cetmix_tower_server/models/cx_tower_command.py
```diff
@@ -140,6 +140,26 @@ class CxTowerCommand(models.Model):
...
+        "- Overwrite: Replace the existing file with the new one.\n"
+        "- Raise Error: Raise an error if the file already exists.",
+    )
+    is_disconnect_from_tmpl = fields.Boolean(
+        string="Disconnect from Template",
+        help=(
+            "If enabled, disconnects the file from its template "
+            "after running the command.\n"
+        ),
+    )
```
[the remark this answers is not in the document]
Kurama2d: In the create_file if_file_exists default value is raise to save previous behavior, but in command itself default value should be skip.

cetmix_tower_server/tests/test_command.py
```diff
@@ -1265,6 +1265,82 @@ def test_run_command_no_command_log(self):
...
+
+        command = self.command_create_file_with_template_server_source
+        command.write({"if_file_exists": "skip"})
+        self.Plan.create(
+            {
+                "name": "Test Flight Plan",
+                "server_ids": [(4, self.server_test_1.id)],
+                "command_ids": [(4, command.id)],
+            }
+        )
```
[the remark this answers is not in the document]
Kurama2d: Thanks. removed

cetmix_tower_server/models/cx_tower_command.py
```diff
@@ -140,6 +140,26 @@ class CxTowerCommand(models.Model):
...
+        "- Skip: Do not create or update the file.\n"
+        "- Overwrite: Replace the existing file with the new one.\n"
+        "- Raise Error: Raise an error if the file already exists.",
+    )
+    is_disconnect_from_tmpl = fields.Boolean(
+        string="Disconnect from Template",
+        help=(
+            "If enabled, disconnects the file from its template "
+            "after running the command.\n",
+        ),
```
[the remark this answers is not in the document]
Kurama2d: Thanks. fixed

cetmix_tower_server/tests/test_command.py
Kurama2d: Added 2 asserts, but for file `code_on_server` field value is taken from the template.

ivs-cetmix (changes requested):
Looks good, however this is still pending:

Please add this new field into the list of exported fields in the [cetmix_tower_yaml](https://github.com/cetmix/cetmix-tower/blob/f5787bee25540a8f9c9269863db1d352a0a45d8d/cetmix_tower_yaml/models/cx_tower_command.py#L10) module. This should be done in a separate commit.

ivs-cetmix:
@Kurama2d please fix the PR name. The mandatory version tag is missing.

cetmix_tower_yaml/tests/test_command.py
```diff
@@ -162,6 +174,16 @@ def test_yaml(command):
                 command.reference,
                 self.command_test_yaml_dict["reference"],
                 "YAML value doesn't match Cetmix Tower one",
+            )
```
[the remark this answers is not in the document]
Kurama2d: Fixed

cetmix_tower_server/models/cx_tower_file_template.py
```diff
@@ -179,5 +179,7 @@ def create_file(self, server, server_dir="", if_file_exists="raise"):
                 "source": self.source,
             }
         )
+        if existing_file and if_file_exists == "overwrite":
+            existing_file.unlink()
```
ivs-cetmix: This may [physically delete](https://github.com/cetmix/cetmix-tower/blob/c653f3bae1e144130c6a741cd5279e0b3b3a89c8/cetmix_tower_server/models/cx_tower_file.py#L299) file from server. Is this what we really want to do?

ivs-cetmix (changes requested):

cetmix_tower_server/models/cx_tower_file_template.py
```diff
@@ -162,12 +162,13 @@ def create_file(self, server, server_dir="", raise_if_exists=False):
...
-        if existing_file:
-            if raise_if_exists:
-                raise ValidationError(_("File already exists on server."))
+        if existing_file and if_file_exists == "skip":
             return existing_file
 
-        return file_model.with_context(is_custom_server_dir=True).create(
+        if existing_file and if_file_exists == "raise":
+            raise ValidationError(_("File already exists on server."))
+
```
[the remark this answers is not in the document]
Kurama2d: I think it's better to have separate conditions for more clear understanding

cetmix_tower_server/models/cx_tower_file_template.py
Kurama2d: It's one of the options, but I decided to change overwrite logic to overwrite data of the existing file.

cetmix_tower_server/tests/test_command.py
```diff
@@ -1265,6 +1265,86 @@ def test_run_command_no_command_log(self):
...
+        new_files = self.server_test_1.file_ids - initial_files
+        self.assertEqual(len(new_files), 1, "Must be one new file created")
+        self.assertEqual(
+            new_files.code_on_server,
+            command.file_template_id.code,
+            "File code must match template",
+        )
+        self.assertFalse(
+            new_files.template_id, "File must be disconnected from template"
+        )
```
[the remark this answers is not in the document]
Kurama2d: It's extra, we have assert for files count above

cetmix_tower_server/tests/test_command.py
```diff
@@ -1265,6 +1265,86 @@ def test_run_command_no_command_log(self):
...
+            orig_file.name,
+            file_template.file_name,
+            "File name must match template after overwrite",
+        )
+        self.assertEqual(
+            orig_file.source,
+            file_template.source,
+            "File source must match template after overwrite",
+        )
+
```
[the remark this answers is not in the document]
Kurama2d: Added

cetmix_tower_server/tests/test_command.py
```diff
@@ -1265,6 +1265,105 @@ def test_run_command_no_command_log(self):
...
+        self.assertEqual(
+            self.env["cx.tower.file"].search_count(
+                [
+                    ("template_id", "=", file_template.id),
+                    ("server_id", "=", self.server_test_1.id),
+                ]
+            ),
+            1,
+            "There must be exactly one physical file record after skip",
+        )
```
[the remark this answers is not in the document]
Kurama2d: Put this assert to it's place after skip part

cetmix_tower_server/tests/test_command.py
```diff
@@ -1265,6 +1265,105 @@ def test_run_command_no_command_log(self):
...
+        new_files = self.server_test_1.file_ids - initial_files
+        self.assertEqual(len(new_files), 1, "Must be one new file created")
+        self.assertEqual(
+            new_files.code_on_server,
+            command.file_template_id.code,
+            "File code must match template",
+        )
+        self.assertFalse(
+            new_files.template_id, "File must be disconnected from template"
+        )
```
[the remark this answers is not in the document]
Kurama2d: It's not needed here

cetmix_tower_server/models/cx_tower_file_template.py
```diff
@@ -162,12 +162,27 @@ def create_file(self, server, server_dir="", raise_if_exists=False):
...
+        if existing_file and if_file_exists == "overwrite":
+            existing_file.with_context(is_custom_server_dir=True).write(
+                {
+                    "name": self.file_name,
+                    "code_on_server": self.code,
```
ivs-cetmix: It's a helper field that is used for manual checks only.
You should use existing functions instead of doing direct writes and manual updates. There is an [example](https://github.com/cetmix/cetmix-tower/blob/0075aa48fe89112d371b9b445f990e588c98f8ac/cetmix_tower_server/models/cx_tower_file_template.py#L115) in the same file.

ivs-cetmix (changes requested):

cetmix_tower_server/models/cx_tower_file_template.py
Kurama2d: Thanks for noticing, replaced with method

cetmix_tower_server/models/cx_tower_file_template.py
```diff
@@ -162,12 +162,21 @@ def create_file(self, server, server_dir="", raise_if_exists=False):
...
-        return file_model.with_context(is_custom_server_dir=True).create(
+        if existing_file and if_file_exists == "raise":
+            raise ValidationError(_("File already exists on server."))
+
+        if existing_file and if_file_exists == "overwrite":
+            existing_file.with_context(is_custom_server_dir=True).write(
+                existing_file._get_file_values_from_related_template()
+            )
+            if existing_file.source == "tower":
+                existing_file.action_push_to_server()
```
ivs-cetmix: Please check the [write](https://github.com/cetmix/cetmix-tower/blob/70e6b9e72ae46431bed8a90495b294352de55d47/cetmix_tower_server/models/cx_tower_file.py#L290) method. The file will be synced automatically if `auto_sync` is enabled, no need to sync it twice.

ivs-cetmix (changes requested):

cetmix_tower_server/tests/test_command.py
```diff
@@ -1265,6 +1265,105 @@ def test_run_command_no_command_log(self):
...
+        # Test if file exists and command is set to "skip"
+        skipped_file = file_template.create_file(
+            server=self.server_test_1,
+            if_file_exists=command.if_file_exists,
+        )
+        self.assertEqual(
+            orig_file,
+            skipped_file,
+            "Skip should return the existing file, not create a new one",
+        )
```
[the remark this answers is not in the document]
Kurama2d: It's not needed to compare id as we compare records

cetmix_tower_server/models/cx_tower_file_template.py
```diff
@@ -162,12 +162,19 @@ def create_file(self, server, server_dir="", raise_if_exists=False):
...
-        return file_model.with_context(is_custom_server_dir=True).create(
+        if existing_file and if_file_exists == "raise":
+            raise ValidationError(_("File already exists on server."))
+
+        if existing_file and if_file_exists == "overwrite":
+            existing_file.with_context(is_custom_server_dir=True).write(
+                existing_file._get_file_values_from_related_template()
+            )
+            return existing_file
+
```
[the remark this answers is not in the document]
Kurama2d: I think it's extra because we can catch error in _get_file_values_from_related_template

cetmix_tower_server/models/cx_tower_file_template.py
Kurama2d: Thanks, removed push to the server

cetmix_tower_server/models/cx_tower_file_template.py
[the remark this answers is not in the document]
ivs-cetmix: @Kurama2d makes sense, could you please check this?

cetmix_tower_server/models/cx_tower_file_template.py
```diff
@@ -162,12 +162,19 @@ def create_file(self, server, server_dir="", raise_if_exists=False):
...
+        if existing_file and if_file_exists == "skip":
             return existing_file
 
-        return file_model.with_context(is_custom_server_dir=True).create(
+        if existing_file and if_file_exists == "raise":
+            raise ValidationError(_("File already exists on server."))
+
+        if existing_file and if_file_exists == "overwrite":
+            existing_file.with_context(is_custom_server_dir=True).write(
+                existing_file._get_file_values_from_related_template()
```
ivs-cetmix: What if existing file is not linked to the same file template or it's not linked to any template at all?
Meanwhile [here](https://github.com/cetmix/cetmix-tower/blob/37b03fb78f9edeea1d93d5f989e954ebf4240785/cetmix_tower_server/models/cx_tower_file_template.py#L158) you are searching for a template linked file only, while not taking into account the case mentioned above.

ivs-cetmix (changes requested):
NB: plese do NOT squash commits until you get the final approval.

cetmix_tower_server/tests/test_command.py
```diff
@@ -1265,6 +1265,109 @@ def test_run_command_no_command_log(self):
...
+                if_file_exists=command.if_file_exists,
+            )
+        # Change command to "overwrite" file if it exists
+        command.write({"if_file_exists": "overwrite"})
+        # Run command again, it should overwrite the file
+        file_template.create_file(
+            server=self.server_test_1,
+            server_dir=file_template.server_dir,
+            if_file_exists=command.if_file_exists,
+        )
```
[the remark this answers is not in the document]
Kurama2d: Below we check that there is only 1 file oh the server, it's the same

cetmix_tower_server/models/cx_tower_file_template.py
```diff
@@ -136,45 +136,66 @@ def create_file(self, server, server_dir="", raise_if_exists=False):
...
+            If the file already exists on the server if_file_exists is set to "raise".
         """
         self.ensure_one()
+        # Explicit guard against invalid behavior values
+        valid_behaviors = {"skip", "raise", "overwrite"}
+        if if_file_exists not in valid_behaviors:
+            raise ValidationError(
+                f"Invalid if_file_exists value: {if_file_exists}. "
+                f"Expected one of {valid_behaviors}."
+            )
```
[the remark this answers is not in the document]
Kurama2d: It's not neccessary here

cetmix_tower_server/models/cx_tower_file_template.py
```diff
@@ -136,45 +136,66 @@ def create_file(self, server, server_dir="", raise_if_exists=False):
...
-            limit=1,
         )
-        if existing_file:
-            if raise_if_exists:
-                raise ValidationError(_("File already exists on server."))
+        # Filter existing files by template if it exists, otherwise take the first one
+        existing_file = (
+            existing_file_ids.filtered(lambda f: f.template_id == self)[:1]
+            or existing_file_ids[:1]
+        )
```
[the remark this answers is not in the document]
Kurama2d: Thanks, added order but not limit because we have template check below

cetmix_tower_server/tests/test_file.py
```diff
@@ -302,18 +302,28 @@ def test_create_file_with_template(self):
...
+        file = file_template.create_file(
+            server=self.server_test_1,
+            server_dir=file_template.server_dir,
+            if_file_exists="overwrite",
+        )
         self.assertEqual(file.code, self.file_template.code)
         self.assertEqual(file.template_id, file_template)
         self.assertEqual(file.server_id, self.server_test_1)
         self.assertEqual(file.source, "tower")
         self.assertEqual(file.server_dir, self.file_template.server_dir)
```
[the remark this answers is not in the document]
Kurama2d: It's not my test so I prefer not to touch anything except something related to my changes

cetmix_tower_server/tests/test_file.py
```diff
@@ -302,18 +302,28 @@ def test_create_file_with_template(self):
...
+                server_dir=file_template.server_dir,
+                if_file_exists="raise",
+            )
 
         another_file = file_template.create_file(
-            server=self.server_test_1, raise_if_exists=False
+            server=self.server_test_1,
+            server_dir=file_template.server_dir,
+            if_file_exists="skip",
         )
```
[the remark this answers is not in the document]
Kurama2d: Not really needed here

cetmix_tower_server/models/cx_tower_file_template.py
```diff
@@ -136,38 +136,55 @@ def create_file(self, server, server_dir="", raise_if_exists=False):
...
```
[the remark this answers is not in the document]
Kurama2d: Added

cetmix_tower_server/models/cx_tower_file_template.py
```diff
@@ -136,38 +136,55 @@ def create_file(self, server, server_dir="", raise_if_exists=False):
...
+        if existing_file and if_file_exists == "skip":
+            return existing_file
+
+        if existing_file and if_file_exists == "raise":
+            raise ValidationError(_("File already exists on server."))
+
+        if existing_file and if_file_exists == "overwrite":
+            existing_file.with_context(is_custom_server_dir=True).write(
+                existing_file._get_file_values_from_related_template()
+            )
```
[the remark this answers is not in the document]
Kurama2d: Yes, replaced to vals

cetmix_tower_server/models/cx_tower_file_template.py
```diff
@@ -136,45 +136,67 @@ def create_file(self, server, server_dir="", raise_if_exists=False):
...
+        existing_file = (
+            existing_file_ids.filtered(lambda f: f.template_id == self)[:1]
+            or existing_file_ids[:1]
+        )
+        if existing_file and if_file_exists == "skip":
+            return existing_file
+
+        if existing_file and if_file_exists == "raise":
+            raise ValidationError(_("File already exists on server."))
+
```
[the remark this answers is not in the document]
Kurama2d: I keep the previous behavir of the method

cetmix_tower_server/tests/test_command.py
```diff
@@ -1265,6 +1265,110 @@ def test_run_command_no_command_log(self):
...
+
+        initial_files = self.server_test_1.file_ids
+        command = self.command_create_file_with_template_server_source
+
+        command.is_disconnect_from_tmpl = True
+        self.server_test_1.run_command(command=command)
+
+        new_files = self.server_test_1.file_ids - initial_files
+        self.assertEqual(len(new_files), 1, "Must be one new file created")
+        self.assertEqual(
```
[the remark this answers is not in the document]
Kurama2d: It's not really needed here

cetmix_tower_server/tests/test_command.py
```diff
@@ -1265,6 +1265,109 @@ def test_run_command_no_command_log(self):
...
+            self.env["cx.tower.file"].search_count(
+                [
+                    ("template_id", "=", file_template.id),
+                    ("server_id", "=", self.server_test_1.id),
+                ]
+            ),
+            1,
+            "There must be exactly one physical file record after skip",
+        )
```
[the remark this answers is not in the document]
Kurama2d: Added

cetmix_tower_server/models/cx_tower_file_template.py
Kurama2d: Thanks, I covered the cases too

cetmix_tower_server/models/cx_tower_file_template.py
```diff
@@ -152,38 +152,51 @@ def create_file(self, server, server_dir="", if_file_exists="raise"):
...
+                f"Invalid if_file_exists value: {if_file_exists}. "
+                f"Expected one of {valid_behaviors}."
+            )
         file_model = self.env["cx.tower.file"]
-        existing_file = file_model.search(
+        existing_file_ids = file_model.search(
             [
-                ("template_id", "=", self.id),
                 ("server_id", "=", server.id),
                 ("server_dir", "=", server_dir or self.server_dir),
```
ivs-cetmix: This will return files all files with the same source that are located in the same directory.
We need to check the `name` field as well to avoid it.

cetmix_tower_server/models/cx_tower_file_template.py
```diff
@@ -152,38 +152,51 @@ def create_file(self, server, server_dir="", if_file_exists="raise"):
...
                 ("source", "=", self.source),
             ],
-            limit=1,
+            order="id DESC",
+        )
+        # Filter existing files by template if it exists, otherwise take the first one
+        existing_file = (
```
ivs-cetmix, line 190: Why should we do an additional sorting if all the files are already retrieved with the query?
Cannot we simply return `existing_file_ids and existing_file_ids[0]`?

cetmix_tower_server/models/cx_tower_file_template.py
```diff
@@ -152,38 +152,51 @@ def create_file(self, server, server_dir="", if_file_exists="raise"):
...
         if existing_file and if_file_exists == "skip":
             return existing_file
 
         if existing_file and if_file_exists == "raise":
             raise ValidationError(_("File already exists on server."))
 
+        vals = {
+            "template_id": self.id,
+            "name": self.file_name,
+            "code_on_server": self.code,
```
ivs-cetmix: Why are you modifying [this field](https://github.com/cetmix/cetmix-tower/blob/7a658a7f3093208aa9c4818047a3a1a3d22fdc13/cetmix_tower_server/models/cx_tower_file.py#L111) instead of `code`?

cetmix_tower_server/models/cx_tower_file_template.py
```diff
@@ -136,45 +136,67 @@ def create_file(self, server, server_dir="", raise_if_exists=False):
...
+        )
+        if existing_file and if_file_exists == "skip":
+            return existing_file
+
+        if existing_file and if_file_exists == "raise":
+            raise ValidationError(_("File already exists on server."))
+
+        vals = {
+            "template_id": self.id,
+            "name": self.file_name,
```
ivs-cetmix, line 210: If the name is the same, does it make any sense to overwrite it?

ivs-cetmix (changes requested):

cetmix_tower_server/models/cx_tower_file_template.py
Kurama2d: After adding to the search domain it doesn't make sense, removed from overwriting

cetmix_tower_server/models/cx_tower_file_template.py
Kurama2d: Thanks for noticing, it's my mistake, changed

cetmix_tower_server/models/cx_tower_file_template.py
Kurama2d: Because in previous logic we searched the file only with particular template. So to save this logic and combine it with cases when there are no template at all I check if there any file with curent template and if now we can take any file

cetmix_tower_server/models/cx_tower_file_template.py
Kurama2d: Thanks, added `name` field to domain

cetmix_tower_server/models/cx_tower_file_template.py
```diff
@@ -136,45 +136,68 @@ def create_file(self, server, server_dir="", raise_if_exists=False):
...
+            raise ValidationError(
+                f"Invalid if_file_exists value: {if_file_exists}. "
+                f"Expected one of {valid_behaviors}."
+            )
         file_model = self.env["cx.tower.file"]
-        existing_file = file_model.search(
+        existing_file_ids = file_model.search(
             [
-                ("template_id", "=", self.id),
+                ("name", "=", self.file_name),
```
ivs-cetmix: Here is another case we need to keep in mind: "name" field can contain variables. So a file with the name value `/opt/test` and another file with the name value `/opt/{{ branch_name}}` rendered with variable value `branch="test"` will end up with the same file but different file names.

ivs-cetmix (changes requested):

cetmix_tower_server/models/cx_tower_file_template.py
```diff
@@ -136,45 +136,71 @@ def create_file(self, server, server_dir="", raise_if_exists=False):
         :param server: recordset
             The server (cx.tower.server) on which the file should be created. This is a
             required parameter.
+        :param if_file_exists: str, optional
+            Defines the behavior if the file already exists on the server.
         :param server_dir: str, optional
             The directory on the server where the file should be created. If not set,
```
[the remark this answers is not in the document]
Kurama2d: All lines lenght checked by pre-commiter

cetmix_tower_server/models/cx_tower_file_template.py
```diff
@@ -136,45 +136,71 @@ def create_file(self, server, server_dir="", raise_if_exists=False):
...
-            limit=1,
+            order="id DESC",
         )
-        if existing_file:
-            if raise_if_exists:
-                raise ValidationError(_("File already exists on server."))
+        # Filter existing files by name or rendered name
+        existing_file_ids = existing_file_ids.filtered(
+            lambda f: f.name in self.file_name or f.rendered_name in self.file_name
+        )
```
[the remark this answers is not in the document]
Kurama2d: You are right, fixed it

cetmix_tower_server/models/cx_tower_file_template.py
Kurama2d: I added another filtered for name and rendered_name, because rendered_name is not storable field and we can't use it in search and there no better decisions here unfortunately.

cetmix_tower_server/models/cx_tower_file_template.py
```diff
@@ -136,45 +136,71 @@ def create_file(self, server, server_dir="", raise_if_exists=False):
...
+            or existing_file_ids[:1]
+        )
+        if existing_file and if_file_exists == "skip":
+            return existing_file
+
+        if existing_file and if_file_exists == "raise":
+            raise ValidationError(_("File already exists on server."))
+
+        vals = {
+            "template_id": self.id,
```
ivs-cetmix, line 213: You can simply add the `template_id` and then call the [onchange method](https://github.com/cetmix/cetmix-tower/blob/aa623db51b2bc44e2d0cefeadd55246ef870b437/cetmix_tower_server/models/cx_tower_file.py#L247). 
This is a better approach because it allows to ensure that no further possible changes will break the code.

ivs-cetmix (changes requested):

cetmix_tower_server/models/cx_tower_file_template.py
Kurama2d: I tried this method and many tests failed for some reason (not only mine ones). I checked data that is set in this method and it's the same as in my logic, but unfortunately something happens. I'm not seure if it worth further checking and changes here.

ivs-cetmix: I think you need to check what exactly fails and why. Probably there are some minor issues that can be easily tweaked. Anyway it's not a good idea to have the same operations spread across several functions in the code. It makes the code error prone and less maintainable.

cetmix_tower_server/models/cx_tower_file_template.py
```diff
@@ -136,45 +136,71 @@ def create_file(self, server, server_dir="", raise_if_exists=False):
...
+        if existing_file and if_file_exists == "raise":
+            raise ValidationError(_("File already exists on server."))
+
+        if existing_file and if_file_exists == "overwrite":
+            existing_file.with_context(is_custom_server_dir=True).write(
+                {
+                    "template_id": self.id,
+                }
+            )
+            return existing_file
```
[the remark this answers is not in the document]
Kurama2d: template_id updates and inverse triggers so we sync data from template to the file.

cetmix_tower_server/models/cx_tower_file_template.py
Kurama2d: Changed to simply write template_id, then _inverse_template_id triggers and updates data (not onchange).

cetmix_tower_server/models/cx_tower_file_template.py
```diff
@@ -136,45 +136,71 @@ def create_file(self, server, server_dir="", raise_if_exists=False):
...
+                f"Invalid if_file_exists value: {if_file_exists}. "
+                f"Expected one of {valid_behaviors}."
+            )
         file_model = self.env["cx.tower.file"]
-        existing_file = file_model.search(
+        existing_file_ids = file_model.search(
             [
-                ("template_id", "=", self.id),
                 ("server_id", "=", server.id),
                 ("server_dir", "=", server_dir or self.server_dir),
```
ivs-cetmix: `server_dir` is also using variables, so using it in domains may lead to an irrelevant result.

cetmix_tower_server/models/cx_tower_file_template.py
```diff
@@ -136,45 +136,71 @@ def create_file(self, server, server_dir="", raise_if_exists=False):
...
             ],
-            limit=1,
+            order="id DESC",
         )
-        if existing_file:
-            if raise_if_exists:
-                raise ValidationError(_("File already exists on server."))
+        # Filter existing files by name or rendered name
+        existing_file_ids = existing_file_ids.filtered(
+            lambda f: f.name == self.file_name or f.rendered_name == self.file_name
```
ivs-cetmix: You should filter by the [full server path](https://github.com/cetmix/cetmix-tower/blob/d1fa7bb2388523bad944ac9679de1bc255b91dd1/cetmix_tower_server/models/cx_tower_file.py#L66) because it contains both render file name and directory.

ivs-cetmix (changes requested):

cetmix_tower_server/models/cx_tower_file_template.py
Kurama2d: Reworked logic with rendered and unrendered server directory and file name

cetmix_tower_server/models/cx_tower_file_template.py
Kurama2d: Reworked logic with rendered and unrendered server directory and file name

cetmix_tower_server/models/cx_tower_file_template.py
```diff
@@ -136,45 +136,94 @@ def create_file(self, server, server_dir="", raise_if_exists=False):
...
+        possible_paths = list(
             {
-                "template_id": self.id,
-                "server_id": server.id,
-                "name": self.file_name,
-                "code_on_server": self.code,
-                "server_dir": server_dir or self.server_dir,
-                "file_type": self.file_type,
-                "source": self.source,
+                path.strip("/") if path.startswith("//") else path
```
ivs-cetmix: What if path starts with `///` or some other special symbol?

I think we need to straighten the check algorithm:
1. Get the path/server_dir and the file name from the template and compose the final URL. 
2. Render them using the `render_code_custom` function like it's done [here](https://github.com/cetmix/cetmix-tower/blob/88464c5841395a3b80aa7832241e7aea2da54ead/cetmix_tower_server/models/cx_tower_file.py#L221).
3.  Compare them with the `full_server_path` values of the records got with search.

ivs-cetmix (changes requested):

cetmix_tower_server/models/cx_tower_file_template.py
Kurama2d: Agree,  changed the logic with rendering path from template

cetmix_tower_server/models/cx_tower_file_template.py
```diff
@@ -136,45 +136,86 @@ def create_file(self, server, server_dir="", raise_if_exists=False):
...
+        # Explicit guard against invalid behavior values
+        valid_behaviors = {"skip", "raise", "overwrite"}
+        if if_file_exists not in valid_behaviors:
+            raise ValidationError(
+                f"Invalid if_file_exists value: {if_file_exists}. "
+                f"Expected one of {valid_behaviors}."
+            )
         file_model = self.env["cx.tower.file"]
-        existing_file = file_model.search(
+        existing_file_ids = file_model.search(
```
ivs-cetmix: `existing_files` because we are fetching records, not ids

ivs-cetmix (changes requested):

cetmix_tower_server/models/cx_tower_file_template.py
Kurama2d: You are right, fixed

ivs-cetmix (approved):
LGTM

CetmixGitDrone:

GabbasovDinar (approved):
LGTM

cetmix_tower_server/models/cx_tower_server.py
```diff
@@ -1232,6 +1228,17 @@ def _command_runner_file_using_template(
...
+            if log_record.command_id.is_disconnect_from_tmpl:
+                file.action_unlink_from_template()
+
+            if log_record.command_id.if_file_exists == "skip":
+                return log_record.finish(
+                    fields.Datetime.now(),
+                    0,
+                    _("File already exists on server. Upload skipped"),
+                    None,
+                )
```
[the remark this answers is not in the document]
Kurama2d: No, it should be after operations. We also have is_disconnect_from_tmpl that can be set so we need to disconect the file even if it was skipped.

CetmixGitDrone:
Congratulations, PR rebased to [14.0](https://github.com/cetmix/cetmix-tower/commits/14.0).

Aldeigja (approved):
Functional LGTM

cetmix_tower_server/models/cx_tower_command.py
```diff
@@ -155,6 +155,26 @@ class CxTowerCommand(models.Model):
...
+            ("overwrite", "Overwrite"),
+            ("raise", "Raise Error"),
+        ],
+        default="skip",
+        help="What to do if file already exists on the server.\n"
+        "- Skip: Do not create or update the file.\n"
+        "- Overwrite: Replace the existing file with the new one.\n"
+        "- Raise Error: Raise an error if the file already exists.",
+    )
+    is_disconnect_from_tmpl = fields.Boolean(
```
ivs-cetmix: Let's rename this field into `disconnect_file` because although `is_disconnect_from_tmpl` is compliant with OCA standards, it's not so user friendly for YAML templates.

ivs-cetmix (changes requested):
Some minor change, otherwise looks good.

cetmix_tower_server/models/cx_tower_command.py
Kurama2d: Renamed.

cetmix_tower_server/models/cx_tower_server.py
```diff
@@ -1220,16 +1216,32 @@ def _command_runner_file_using_template(
...
+            if log_record.command_id.disconnect_file:
+                file.action_unlink_from_template()
+
+            if is_creation_skipped:
+                return log_record.finish(
+                    fields.Datetime.now(),
+                    0,
+                    _("File already exists on server. Upload skipped"),
+                    None,
                 )
```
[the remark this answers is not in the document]
Kurama2d: No changes needed anymore.

ivs-cetmix (approved):
/ocabot merge minor

CetmixGitDrone:
Prepared branch [14.0-ocabot-merge-pr-309-by-ivs-cetmix-bump-minor](https://github.com/cetmix/cetmix-tower/commits/14.0-ocabot-merge-pr-309-by-ivs-cetmix-bump-minor), awaiting test results.

CetmixGitDrone:
Congratulations, your PR was merged at 130bd169f7642aeb03b13d5ade78d98884fab2a0. Thanks a lot for contributing to cetmix. ❤️

PR #347, opened 2025-07-14
[16.0][ADD] cetmix_tower_server-implement_vault_for_secrets

halbtonjazz:
Implement a new approach for storing confidential data in a centralized
way to avoid code duplication and improve maintainability.

Currently we have several occasions in several models where we store 
confidential information. This means that we have the same code 
duplicated several times which makes code maintenance less convenient 
and is also error prone.

Changes include:
- Add new abstract model cx.tower.vault.mixin to handle secret values
- Add new model cx.tower.vault to store the secret values  
- Apply the mixin to cx.tower.key, cx.tower.key.value, and
  cx.tower.server models
- Move secret handling logic from individual models to the mixin
- Set proper security rules for the vault model

This is a draft implementation that provides centralized secret storage
mechanism which can be easily extended to use external key management
systems by overriding the mixin functions.

Task: 4824

## Summary by CodeRabbit

- New Features
  - Secure, vault-backed storage for sensitive fields (passwords, keys) across servers and keys.
  - Secrets are redacted with a placeholder in views and command outputs, with controlled retrieval when needed.
  - Creation-time validation ensures SSH password is provided when using password-based authentication.

- UI
  - Broader menu visibility by removing group restrictions from key sections.
  - Tools menu repositioned under the root menu; Settings order adjusted.

### DIFF
--- a/cetmix_tower_server/models/cx_tower_key.py
+++ b/cetmix_tower_server/models/cx_tower_key.py
@@ -9,13 +9,16 @@ class CxTowerKey(models.Model):
 
     _name = "cx.tower.key"
     _description = "Cetmix Tower Key/Secret Storage"
-    _inherit = ["cx.tower.reference.mixin", "cx.tower.access.role.mixin"]
+    _inherit = [
+        "cx.tower.reference.mixin",
+        "cx.tower.access.role.mixin",
+        "cx.tower.vault.mixin",
+    ]
     _order = "name"
 
     KEY_PREFIX = "#!cxtower"
     KEY_TERMINATOR = "!#"
-    SECRET_VALUE_PLACEHOLDER = "*** Insert new value to replace the existing one ***"
-    SECRET_VALUE_SPOILER = "*****"
+    SECRET_FIELDS = ["secret_value"]
 
     key_type = fields.Selection(
         selection=[
@@ -71,40 +74,6 @@ class CxTowerKey(models.Model):
             else:
                 rec.reference_code = None
 
-    @api.model_create_multi
-    def create(self, vals_list):
-        """Create a new key and set secret value if key type is secret"""
-
-        # Create records one by one to ensure that we preserve
-        # the original value list order.
-        # This is done to avoid possible errors when this list is modified
-        # by some other module that overrides this function.
-        # We also need to avoid secret values being written to the database
-        # any other way besides the _set_secret_value method.
-        records = self.browse()
-        for vals in vals_list:
-            secret_value = vals.pop("secret_value", None)
-            new_key = super().create(vals)
-            if secret_value:
-                new_key._set_secret_value(secret_value)
-            records |= new_key
-        return records
-
-    def write(self, vals):
-        """Write key values"""
-        # Remove secret value from vals and set it later
-        if "secret_value" in vals and not self.env.context.get("write_secret_value"):
-            secret_value = vals.pop("secret_value", None)
-            has_secret_value = True
-        else:
-            has_secret_value = False
-        res = super().write(vals)
-        # Set secret value
-        if has_secret_value:
-            for key in self:
-                key._set_secret_value(secret_value)
-        return res
-
     def _get_reference_pattern(self):
         """
         Override mixin method
@@ -130,19 +99,6 @@ class CxTowerKey(models.Model):
         else:
             key_prefix = None
         return key_prefix
-
-    def _read(self, fields):  # pylint: disable=missing-return # doesn't return anything
-        """Substitute fields based on api"""
-        super()._read(fields)
-        if "secret_value" in fields or fields == []:
-            # Public user used for substitution
-            for record in self:
-                try:
-                    record._cache["secret_value"] = self.SECRET_VALUE_PLACEHOLDER
-                except Exception:  # pylint: disable=except-pass
-                    # skip SpecialValue
-                    # (e.g. for missing record or access right)
-                    pass
 
     def _parse_code_and_return_key_values(self, code, pythonic_mode=False, **kwargs):
         """Replaces key placeholders in code with the corresponding values,
@@ -372,7 +328,7 @@ class CxTowerKey(models.Model):
                 key_value = filtered_key_values[0]
 
         if key_value:
-            return key_value._get_secret_value()
+            return key_value._get_secret_value("secret_value")
 
     def _replace_with_spoiler(self, code, key_values):
         """Helper function that replaces clean text keys in code with spoiler.
@@ -401,40 +357,17 @@ class CxTowerKey(models.Model):
             # If key_value contains an escaped line break replace then remove escaping
             key_value = key_value.replace("\\n", "\n")
             # Replace key including key terminator
-            code = code.replace(key_value, self.SECRET_VALUE_SPOILER)
+            code = code.replace(key_value, self.SECRET_VALUE_PLACEHOLDER)
 
         return code
 
-    def _get_secret_value(self):
-        """Get secret value.
-        Override this method in case you need
-        to implement custom key storages.
-
-        Returns:
-            str: secret value or None if no secret value is found
-        """
-
-        # Return None in case of empty recordset
-        self.ensure_one()
-        self.env.cr.execute(
-            """
-            SELECT secret_value
-            FROM cx_tower_key
-            WHERE id = %s
-            """,
-            [self.id],
-        )
-        result = self.env.cr.fetchone()
-        if result:
-            return result[0]
-
-    def _set_secret_value(self, value=None):
+    def _set_secret_values(self, vals):
         """Set secret value.
         Override this method in case you need
         to implement custom key storages.
 
         Args:
-            value (str): secret value
+            vals (dict): Dictionary of field names to secret values
         """
         self.ensure_one()
         if self.key_type == "s":
@@ -443,15 +376,11 @@ class CxTowerKey(models.Model):
                 lambda x: not x.server_id and not x.partner_id
             )
             if general_value:
-                general_value._set_secret_value(value)
+                general_value._set_secret_values(vals)
             else:
-                self.value_ids.create(
-                    {
-                        "key_id": self.id,
-                        "secret_value": value,
-                    }
-                )
+                create_vals = {"key_id": self.id}
+                create_vals.update(vals)
+                self.value_ids.create(create_vals)
 
         elif self.key_type == "k":
-            self.with_context(write_secret_value=True).write({"secret_value": value})
-            self.invalidate_recordset(["secret_value"])
+            return super()._set_secret_values(vals)
--- a/cetmix_tower_server/models/cx_tower_server.py
+++ b/cetmix_tower_server/models/cx_tower_server.py
@@ -83,9 +83,12 @@ class CxTowerServer(models.Model):
         "cx.tower.reference.mixin",
         "mail.thread",
         "mail.activity.mixin",
+        "cx.tower.vault.mixin",
     ]
     _description = "Cetmix Tower Server"
     _order = "name asc"
+
+    SECRET_FIELDS = ["ssh_password", "host_key"]
 
     # ---- Main
     active = fields.Boolean(default=True)
@@ -124,7 +127,8 @@ class CxTowerServer(models.Model):
         string="SSH Username", required=True, groups="cetmix_tower_server.group_manager"
     )
     ssh_password = fields.Char(
-        string="SSH Password", groups="cetmix_tower_server.group_manager"
+        string="SSH Password",
+        groups="cetmix_tower_server.group_manager",
     )
     ssh_key_id = fields.Many2one(
         comodel_name="cx.tower.key",
@@ -286,7 +290,6 @@ class CxTowerServer(models.Model):
         """Ensure SSH settings are valid.
         Set 'skip_ssh_settings_check' context key to skip the checks
         """
-
         # Skip the check if context key is set
         if self._context.get("skip_ssh_settings_check"):
             return
@@ -301,10 +304,6 @@ class CxTowerServer(models.Model):
                         srv=rec.name,
                     )
                 )
-            if rec.ssh_auth_mode == "p" and not rec.ssh_password:
-                validation_errors.append(
-                    _("Please provide SSH password for %(srv)s", srv=rec.name)
-                )
             if rec.ssh_auth_mode == "k" and not rec.ssh_key_id:
                 validation_errors.append(
                     _("Please provide SSH Key for %(srv)s", srv=rec.name)
@@ -314,13 +313,6 @@ class CxTowerServer(models.Model):
             if validation_errors:
                 validation_error = "\n".join(validation_errors)
                 raise ValidationError(validation_error)
-
-    def write(self, vals):
-        """Invalidate host_key cache"""
-        res = super().write(vals)
-        if "host_key" in vals:
-            self.invalidate_recordset(["host_key"])
-        return res
 
     def unlink(self):
         """Run post-delete flight plan"""
@@ -359,22 +351,6 @@ class CxTowerServer(models.Model):
                 server.status = "deleting"
 
         return super(CxTowerServer, servers_to_delete).unlink()
-
-    def _read(self, fields):  # pylint: disable=missing-return # doesn't return anything
-        """Replace host_key with secret value spoiler"""
-        super()._read(fields)
-        spoiler = self.env["cx.tower.key"].SECRET_VALUE_SPOILER
-        # Public user used for substitution
-        if "host_key" in fields or fields == []:
-            for record in self:
-                try:
-                    record._cache["host_key"] = (
-                        spoiler if record._cache["host_key"] else None
-                    )
-                except Exception:  # pylint: disable=except-pass
-                    # skip SpecialValue
-                    # (e.g. for missing record or access right)
-                    pass
 
     @api.returns("self", lambda value: value.id)
     def copy(self, default=None):
@@ -551,7 +527,7 @@ class CxTowerServer(models.Model):
         self.ensure_one()
         self = self.sudo()
         try:
-            host_key = self._get_host_key_value()
+            host_key = self._get_secret_value("host_key")
 
             # Check host only if IP address is present
             skip_host_key = skip_host_key or self.skip_host_key
@@ -733,7 +709,7 @@ class CxTowerServer(models.Model):
             Char: password ready to be used for connection parameters
         """
         self.ensure_one()
-        password = self.sudo().ssh_password
+        password = self._get_secret_value("ssh_password")
         return password
 
     def _get_ssh_key(self):
@@ -748,31 +724,10 @@ class CxTowerServer(models.Model):
         # regardless of access rights
         if self.sudo().ssh_key_id:
             # Use context key to read secret value
-            ssh_key = self.ssh_key_id._get_secret_value()
+            ssh_key = self.ssh_key_id._get_secret_value("secret_value")
         else:
             ssh_key = None
         return ssh_key
-
-    def _get_host_key_value(self):
-        """Get host key value
-
-        Returns:
-            Char: Host key value
-        """
-        # Return None in case of empty recordset
-        if not self:
-            return
-        self.env.cr.execute(
-            """
-            SELECT host_key
-            FROM cx_tower_server
-            WHERE id = %s
-            """,
-            [self.id],
-        )
-        result = self.env.cr.fetchone()
-        if result:
-            return result[0]
 
     @ensure_ssh_disconnect
     def _get_host_key_from_host(self, raise_on_error=True, timeout=60):
@@ -1980,3 +1935,21 @@ class CxTowerServer(models.Model):
                 "sticky": sticky,
             },
         }
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        """Override create to validate SSH password before record creation."""
+        # Validate SSH password before creating records
+        if not self._context.get("skip_ssh_settings_check"):
+            validation_errors = []
+            for vals in vals_list:
+                if vals.get("ssh_auth_mode") == "p" and not vals.get("ssh_password"):
+                    server_name = vals["name"]
+                    validation_errors.append(
+                        _("Please provide SSH password for %(srv)s", srv=server_name)
+                    )
+
+            if validation_errors:
+                raise ValidationError("\n".join(validation_errors))
+
+        return super().create(vals_list)
--- a/cetmix_tower_server/models/cx_tower_vault.py
+++ b/cetmix_tower_server/models/cx_tower_vault.py
@@ -1,0 +1,36 @@
+from odoo import fields, models
+
+
+class CxTowerVault(models.Model):
+    _name = "cx.tower.vault"
+    _description = "Cetmix Tower Vault"
+
+    res_model = fields.Char(
+        string="Resource Model",
+        required=True,
+        help="Model name of the resource that uses this vault",
+    )
+    res_id = fields.Many2oneReference(
+        string="Resource ID",
+        model_field="res_model",
+        help="ID of the resource that uses this vault",
+        required=True,
+    )
+    field_name = fields.Char(
+        required=True,
+        help="Name of the field that contains the secret value",
+    )
+    data = fields.Text(
+        string="Secret Data",
+        required=True,
+        copy=False,
+        help="The secret data to be stored in the vault",
+    )
+
+    _sql_constraints = [
+        (
+            "vault_unique_key",
+            "UNIQUE(res_model, res_id, field_name)",
+            "Each secret (model, record, field) must be unique in the vault.",
+        ),
+    ]
--- a/cetmix_tower_server/models/cx_tower_vault_mixin.py
+++ b/cetmix_tower_server/models/cx_tower_vault_mixin.py
@@ -1,0 +1,392 @@
+from collections import defaultdict
+
+from odoo import api, models
+
+
+class CxTowerVaultMixin(models.AbstractModel):
+    """Mixin for vault functionality.
+
+    This mixin provides methods to securely store and retrieve sensitive data
+    in the vault. Inheriting models must define SECRET_FIELDS list with field
+    names that should be stored in the vault.
+    """
+
+    _name = "cx.tower.vault.mixin"
+    _description = "Cetmix Tower Vault Mixin"
+
+    SECRET_VALUE_PLACEHOLDER = "*****"
+    SECRET_FIELDS = []
+
+    def _read(self, fields):  # pylint: disable=missing-return # doesn't return anything
+        """Substitute fields based on api.
+
+        This method replaces values of secret fields with a placeholder value
+        when they are read from the database.
+
+        Args:
+            fields (list): List of fields to read
+        """
+        super()._read(fields)
+
+        show_all = not fields
+        secret_fields = (
+            self.SECRET_FIELDS
+            if show_all
+            else [f for f in self.SECRET_FIELDS if f in fields]
+        )
+
+        for record in self:
+            for secret_field in secret_fields:
+                try:
+                    record._cache[secret_field] = self.SECRET_VALUE_PLACEHOLDER
+                except Exception:  # pylint: disable=except-pass
+                    # skip SpecialValue
+                    # (e.g. for missing record or access right)
+                    pass
+
+    def _get_secret_value(self, field_name):
+        """Retrieves the actual secret value for a specific field for a single record.
+
+        This method is the only way to get the real secret field value because:
+        - Direct field access (e.g., self.secret_field)
+          returns placeholder due to _read() override
+        - The actual field in the main table is empty/NULL
+          as values are stored in vault
+
+        Args:
+            field_name (str): Name of the secret field to retrieve
+
+        Returns:
+            str or None: The actual secret value, or None if not found or field
+                        is not in SECRET_FIELDS
+
+        Note:
+            This method bypasses Odoo's ORM field access to avoid getting
+            placeholder values returned by the overridden _read() method.
+        """
+
+        self.ensure_one()
+
+        return self._get_secret_values([field_name]).get(self.id, {}).get(field_name)
+
+    def _get_secret_values(self, fields_list=None):
+        """Retrieve secret values from the vault for specified fields.
+
+        This method fetches secret values stored in the vault for all records
+        in the current recordset and specified fields (or all SECRET_FIELDS).
+
+        Args:
+            fields_list (list, optional): List of field names to retrieve.
+                                        Defaults to all SECRET_FIELDS.
+
+        Returns:
+            dict: Dictionary mapping record IDs to their secret field values.
+                  Structure: {res_id: {field_name: secret_value}}
+
+                  Example:
+                  {1: {'ssh_password': 'secret123', 'host_key': 'key456'},
+                   2: {'ssh_password': 'secret789'}}
+
+        Note:
+            This method searches vault records using standard domain filtering
+            by res_id, and field_name for reliable record matching.
+        """
+
+        fields_to_fetch = fields_list or self.SECRET_FIELDS
+        if not (self and fields_to_fetch):
+            return {}
+
+        # Search vault records for all records and all secret fields
+        domain = [
+            ("res_model", "=", self._name),
+            ("res_id", "in", self.ids),
+            ("field_name", "in", fields_to_fetch),
+        ]
+        vault_records = (
+            self.env["cx.tower.vault"]
+            .sudo()
+            .search_read(
+                domain,
+                ["res_id", "field_name", "data"],
+            )
+        )
+        res = defaultdict(dict)
+        for record in vault_records:
+            res[record["res_id"]][record["field_name"]] = record["data"]
+
+        return dict(res)
+
+    def _set_secret_values(self, vals):
+        """Store secret values in the vault.
+
+        This method stores sensitive data in the vault for all records in the recordset.
+        It either updates existing vault records or creates new ones for each
+        record-field pair in the vals dictionary.
+
+        This method can be overridden to implement custom storage mechanisms
+        for secret values, such as external key management systems or
+        encryption services.
+
+        Args:
+            vals (dict): Dictionary mapping field names to their secret values
+                         to be stored in the vault for all records
+
+        Returns:
+            None
+        """
+        if not vals or not self:
+            return
+
+        # Get all existing vault records in ONE SQL query
+        domain = [
+            ("res_model", "=", self._name),
+            ("res_id", "in", self.ids),
+            ("field_name", "in", list(vals.keys())),
+        ]
+        existing_vault_records = self.env["cx.tower.vault"].sudo().search(domain)
+
+        # Prepare data for batch operations
+        vals_to_update_records = defaultdict(lambda: self.env["cx.tower.vault"])
+        records_to_unlink = self.env["cx.tower.vault"]
+        records_to_create = []
+
+        # Process each record and field combination
+        for record in self:
+            for field, value in vals.items():
+                # Find existing vault record using filtered (in-memory operation)
+                existing_record = existing_vault_records.filtered(
+                    lambda v, record_id=record.id, field_name=field: (
+                        v.res_id == record_id and v.field_name == field_name
+                    )
+                )
+
+                if existing_record:
+                    if value:
+                        vals_to_update_records[value] |= existing_record
+
+                    else:
+                        records_to_unlink |= existing_record
+
+                else:
+                    if not value:
+                        continue
+
+                    records_to_create.append(
+                        {
+                            "res_model": self._name,
+                            "res_id": record.id,
+                            "field_name": field,
+                            "data": value,
+                        }
+                    )
+
+        # Batch operations
+        for value, records in vals_to_update_records.items():
+            records.sudo().write({"data": value})
+
+        if records_to_create:
+            self.env["cx.tower.vault"].sudo().create(records_to_create)
+        if records_to_unlink:
+            records_to_unlink.sudo().unlink()
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        """Override create to handle secret values securely.
+
+        Extracts secret fields, stores them in vault, and prevents
+        actual secret values from being saved in the main table.
+
+        Args:
+            vals_list (list): List of dictionaries containing field values
+                             for record creation
+
+        Returns:
+            recordset: Created records with secret values stored in vault
+
+        Note:
+            Secret fields are automatically processed and stored securely.
+            The main database table never contains actual secret values.
+        """
+
+        # Step 1: Extract secret fields and generate temporary IDs
+        secret_vals = self._extract_and_replace_secret_fields(vals_list)
+
+        # Step 2: Create records with batch operation
+        records = super().create(vals_list)
+
+        # Step 3: Update vault records with real IDs
+        if secret_vals:
+            self._process_secret_values_after_creation(records, secret_vals)
+
+        return records
+
+    def write(self, vals):
+        """Override write to handle secret fields.
+
+        Extracts secret field values from vals dictionary and stores them securely
+        in the vault instead of the main database table. The remaining non-secret
+        fields are processed by the standard write method.
+
+        Args:
+            vals (dict): Dictionary of field values to write to records
+
+        Returns:
+            bool: Result of the parent write operation
+
+        Note:
+            Secret fields defined in SECRET_FIELDS are automatically intercepted
+            and stored in vault. Cache is invalidated for all secret fields when
+            any secret field is modified.
+        """
+        # Extract secret fields
+        secret_values = {}
+        for secret_field in self.SECRET_FIELDS:
+            if secret_field in vals:
+                secret_values[secret_field] = vals.pop(secret_field)
+
+        res = super().write(vals)
+
+        if secret_values:
+            self._set_secret_values(secret_values)
+            # Invalidate cache for all secret fields
+            self.invalidate_recordset(self.SECRET_FIELDS)
+
+        return res
+
+    def unlink(self):
+        """Override unlink to delete vault records.
+
+        Automatically removes all associated vault records before deleting
+        the main records to prevent orphaned secret data in the vault.
+
+        Returns:
+            bool: Result of the parent unlink operation
+
+        Note:
+            Vault cleanup is performed automatically and cannot be bypassed.
+        """
+        ids = self.ids
+
+        res = super().unlink()
+
+        # Find all vault records for these records
+        vault_records = (
+            self.env["cx.tower.vault"]
+            .sudo()
+            .search([("res_model", "=", self._name), ("res_id", "in", ids)])
+        )
+
+        # Delete vault records
+        if vault_records:
+            vault_records.sudo().unlink()
+
+        return res
+
+    def _extract_and_replace_secret_fields(self, vals_list):
+        """Extract secret fields and replace with temporary identifiers.
+
+        Processes value dictionaries for record creation, replacing secret field values
+        with unique temporary identifiers. The actual secret values are mapped to these
+        temporary identifiers for later secure storage in the vault system.
+
+        Args:
+            vals_list (list): List of value dictionaries for record creation.
+
+        Returns:
+            tuple: (secret_vals dict mapping temp IDs to secret values,
+                    modified vals_list with temp identifiers)
+
+        Note:
+            Used during record creation as part of the secure secret storage workflow.
+        """
+        temp_id_counter = 0
+        secret_vals = {}
+
+        for vals in vals_list:
+            for secret_field in self.SECRET_FIELDS:
+                if secret_field in vals and vals[secret_field]:
+                    temp_id_counter += 1
+                    temp_identifier = str(temp_id_counter)
+                    secret_vals[temp_identifier] = vals[secret_field]
+                    vals[secret_field] = temp_identifier
+
+        return secret_vals
+
+    def _process_secret_values_after_creation(self, records, secret_vals):
+        """Process secret values after records creation.
+
+        Replaces temporary identifiers with actual secret values in the vault
+        and invalidates cache for affected fields.
+
+        Args:
+            records (recordset): Newly created records with temporary identifiers
+            secret_vals (dict): Mapping of temporary identifiers to secret values
+
+        Returns:
+            None
+
+        Note:
+            Called automatically during create() process. Should not be used directly.
+        """
+        fields_str = ", ".join(self.SECRET_FIELDS)
+        query = f"SELECT id, {fields_str} FROM {self._table} WHERE id in %s"
+        self.env.cr.execute(query, (tuple(records.ids),))
+        records_dict = self.env.cr.dictfetchall()
+
+        for record_dict in records_dict:
+            self._process_single_record_secrets(record_dict, secret_vals)
+
+        records._clear_temp_values()
+        records.invalidate_recordset(self.SECRET_FIELDS)
+
+    def _process_single_record_secrets(self, record_dict, secret_vals):
+        """Process secrets for a single record.
+
+        Replaces temporary identifiers with actual secret values for one record,
+        clears temporary values from main table and stores secrets in vault.
+
+        Args:
+            record_dict (dict): Dictionary with record data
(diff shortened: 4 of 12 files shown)


cetmix_tower_server/models/cx_tower_vault_mixin.py
```diff
@@ -0,0 +1,191 @@
...
+        Args:
+            fields (list): List of field names to retrieve secret values for
+
+        Returns:
+            list: List of dictionaries containing the retrieved vault records with
+                  their data field. Each dictionary represents a vault record.
+        """
+        domain = [
+            ("res_model", "=", self._name),
+            ("res_id", "=", self.id),
```
ivs-cetmix: What if `self` contains multiple records?

cetmix_tower_server/models/cx_tower_vault_mixin.py
```diff
@@ -0,0 +1,191 @@
...
+        if not vals:
+            return
+
+        for field, value in vals.items():
+            if not value:
+                continue
+
+            domain = [
+                ("res_model", "=", self._name),
+                ("res_id", "=", self.id),
```
ivs-cetmix: Same here

cetmix_tower_server/models/cx_tower_vault_mixin.py
```diff
@@ -0,0 +1,191 @@
...
+                ("field_name", "=", field),
+            ]
+            vaults_data = self.env["cx.tower.vault"].sudo().search(domain)
```
ivs-cetmix: Running SQL query inside a loop is really a bad idea in terms of performance. You should fetch all the data outside the look and use helper functions such as `filtered` afterwards.

cetmix_tower_server/models/cx_tower_vault_mixin.py
```diff
@@ -0,0 +1,191 @@
...
+
+            if vaults_data:
+                vaults_data.write({"data": value})
+            else:
+                self.env["cx.tower.vault"].sudo().create(
```
ivs-cetmix: Same for `create`. This function supports val_list, so you should prepare a list inside the loop and than call the `create` once.

cetmix_tower_server/models/cx_tower_vault_mixin.py
```diff
@@ -0,0 +1,191 @@
...
+        records = self.browse()
+        for vals in vals_list:
+            # Extract secret fields
+            secret_values = {}
+            for secret_field in self.SECRET_FIELDS:
+                if secret_field in vals:
+                    secret_values[secret_field] = vals.pop(secret_field)
+
+            # Create record without secret fields
+            new_record = super().create(vals)
```
ivs-cetmix: Why not to sanitise the vals and create all the records afterwards?

ivs-cetmix (changes requested):

cetmix_tower_server/models/cx_tower_vault_mixin.py
halbtonjazz: I have resolved this issue in a new commit. Please check it

cetmix_tower_server/models/cx_tower_vault_mixin.py
halbtonjazz: I have resolved this issue in a new commit. Please check it

cetmix_tower_server/models/cx_tower_vault_mixin.py
halbtonjazz: Please review the update

cetmix_tower_server/models/cx_tower_vault_mixin.py
halbtonjazz: Please review the update

cetmix_tower_server/models/cx_tower_vault_mixin.py
halbtonjazz: Yes, you're right, thanks for the suggestion

cetmix_tower_server/models/cx_tower_vault_mixin.py
```diff
@@ -0,0 +1,230 @@
...
+    names that should be stored in the vault.
+    """
+
+    _name = "cx.tower.vault.mixin"
+    _description = "Cetmix Tower Vault Mixin"
+
+    SECRET_VALUE_PLACEHOLDER = "*** Insert new value to replace the existing one ***"
+    SECRET_VALUE_SPOILER = "*****"
+    SECRET_FIELDS = []
+    _placeholder_field = "SECRET_VALUE_PLACEHOLDER"
```
ivs-cetmix: Why do we need to add an extra variable instead of simply using the `SECRET_VALUE_PLACEHOLDER`?

cetmix_tower_server/models/cx_tower_vault_mixin.py
```diff
@@ -0,0 +1,230 @@
...
+        domain = [
+            ("res_model", "=", self._name),
+            ("res_id", "in", self.ids),
+            ("field_name", "in", field_list),
+        ]
+        vault_records = self.env["cx.tower.vault"].sudo().search(domain)
+
+        # Group vault data by res_id
+        vault_data_by_record = {}
+        for vault_record in vault_records:
```
ivs-cetmix: Why not simply to use [search_read](https://github.com/odoo/odoo/blob/c637426f676c8fd142196fc8610995ebe9ef82b1/odoo/models.py#L5038) instead?

cetmix_tower_server/models/cx_tower_vault_mixin.py
```diff
@@ -0,0 +1,230 @@
...
+        # Get all existing vault records in ONE SQL query
+        domain = [
+            ("res_model", "=", self._name),
+            ("res_id", "in", self.ids),
+            ("field_name", "in", list(vals.keys())),
+        ]
+        existing_vault_records = self.env["cx.tower.vault"].sudo().search(domain)
+
+        # Prepare data for batch operations
+        records_to_update = []
```
ivs-cetmix: You should opt for using reocrdset when working with recordsets, not a list.

cetmix_tower_server/models/cx_tower_vault_mixin.py
```diff
@@ -0,0 +1,230 @@
...
+                # Find existing vault record using filtered (in-memory operation)
+                existing_record = existing_vault_records.filtered(
+                    lambda v, record_id=record.id, field_name=field: v.res_id
+                    == record_id
+                    and v.field_name == field_name
+                )
+
+                if existing_record:
+                    # Prepare for update
+                    records_to_update.append((existing_record, str(value)))
```
ivs-cetmix: Check the comment above

cetmix_tower_server/models/cx_tower_vault_mixin.py
```diff
@@ -0,0 +1,230 @@
...
+                        {
+                            "res_model": self._name,
+                            "res_id": record.id,
+                            "field_name": field,
+                            "data": str(value),
+                        }
+                    )
+
+        # Batch operations
+        for vault_record, new_value in records_to_update:
```
ivs-cetmix: Can it happen so that different records will have different values? If no, then we need to update all records at once.

cetmix_tower_server/models/cx_tower_vault_mixin.py
```diff
@@ -0,0 +1,230 @@
...
+                    secret_values[field] = sanitized_vals.pop(field)
+
+            sanitized_vals_list.append(sanitized_vals)
+            secret_values_list.append(secret_values or None)
+
+        # Batch create all records
+        records = super().create(sanitized_vals_list)
+
+        # Assign secret values to corresponding records
+        for record, secret_values in zip(records, secret_values_list, strict=True):
```
ivs-cetmix: strict doesnt' work on Python 3.10 which is used in Odoo official docker image for 16.0 
Check [this](https://github.com/cetmix/cetmix-tower/blob/c7b1a8fb76090f2d04bcdb1e7809cb23cf0b1683/cetmix_tower_server/tests/test_server.py#L273)

cetmix_tower_server/models/cx_tower_vault_mixin.py
```diff
@@ -0,0 +1,230 @@
...
+        for record in self:
+            if record.id in vault_data_by_record:
+                vaults_data.append(vault_data_by_record[record.id])
+            else:
+                # Record has no vault data, return just id
+                vaults_data.append({"id": record.id})
+
+        return vaults_data
+
+    def _set_secret_value(self, vals):
```
ivs-cetmix: Use plural to reflect the function nature -> `values`

ivs-cetmix (changes requested):

cetmix_tower_server/models/cx_tower_vault_mixin.py
halbtonjazz: Because depending on the model, we need to use `SECRET_VALUE_SPOILER` or `SECRET_VALUE_PLACEHOLDER `(in the `_read` method, as in the current code)

cetmix_tower_server/models/cx_tower_vault_mixin.py
halbtonjazz: fixed

cetmix_tower_server/models/cx_tower_vault_mixin.py
halbtonjazz: The problem is that we cannot compare the created records with the original vals because there is no identifier that can link them together, and the sorting in the model where cx.tower.vault.mixin will be inherited may be different.

cetmix_tower_server/models/cx_tower_vault_mixin.py
halbtonjazz: I had to use `strict=True` because pre-commit was requiring it. However, I've now rewritten the  `create` method to use `@api.model_create_single` instead of `@api.model_create_multi` since single record creation is required because batch creation doesn't allow matching values from  `vals_list` with the created records when storing secret fields in the vault. We need the exact record ID to create corresponding vault entries for each secret field

cetmix_tower_server/models/cx_tower_vault_mixin.py
halbtonjazz: fixed

cetmix_tower_server/models/cx_tower_vault_mixin.py
halbtonjazz: fixed

cetmix_tower_server/models/cx_tower_vault_mixin.py
halbtonjazz: fixed

cetmix_tower_server/models/cx_tower_vault_mixin.py
halbtonjazz: fixed

GabbasovDinar (approved):

GabbasovDinar (changes requested):
@halbtonjazz please, fix tests

cetmix_tower_server/models/cx_tower_vault_mixin.py
```diff
@@ -0,0 +1,269 @@
...
+            if secret_field in vals:
+                secret_values[secret_field] = vals.get(secret_field)
+
+        # Create record without secret fields
+        res = super().create(vals)
+
+        # Store secret values in vault
+        if secret_values:
+            res.set_secret_values(secret_values)
+            table_name = res._table
```
ivs-cetmix: Res is the same model as self, so `self._table`

ivs-cetmix: You should update all the records at once, it will be just a single query instead of multiple ones.

cetmix_tower_server/models/cx_tower_vault_mixin.py
```diff
@@ -0,0 +1,269 @@
...
+            record_id = res.id
+
+            secret_field_list = list(secret_values.keys())
+            for secret_field in secret_field_list:
+                query = f"UPDATE {table_name} SET {secret_field} = NULL WHERE id = %s"
```
ivs-cetmix: Please add a complete comment that describes why you are updating the database table directly here.

ivs-cetmix (changes requested):

cetmix_tower_server/models/cx_tower_vault_mixin.py
halbtonjazz: fixed

cetmix_tower_server/models/cx_tower_vault_mixin.py
halbtonjazz: added

cetmix_tower_server/models/cx_tower_vault_mixin.py
ivs-cetmix: These are two different things used in two different scenarios. Placeholder is always a placeholder, while spoiler is always a spoiler.
Moreover we can simply use only placeholder instead of having two different entities that are doing virtually the same thing.

cetmix_tower_server/models/cx_tower_vault_mixin.py
```diff
@@ -0,0 +1,276 @@
...
+
+        Note:
+            This method bypasses Odoo's ORM field access to avoid getting
+            placeholder values returned by the overridden _read() method.
+        """
+
+        self.ensure_one()
+
+        # Validate field is configured as secret field
+        if field_name not in self.SECRET_FIELDS:
```
ivs-cetmix: Why not to simply call the `get_secrect_values` function and returns a singe result for the selected record?

cetmix_tower_server/models/cx_tower_vault_mixin.py
```diff
@@ -0,0 +1,276 @@
...
+            records.sudo().write({"data": value})
+
+        self.env["cx.tower.vault"].sudo().create(records_to_create)
+        records_to_unlink.sudo().unlink()
+
+    # Single record creation is required because batch creation doesn't allow
+    # matching values from vals_list with the created records when storing
+    # secret fields in the vault. We need the exact record ID to create
+    # corresponding vault entries for each secret field.
+    @api.model_create_single
```
ivs-cetmix: We must avoid single record creation, so must be redesigned to use the `create_multi`.  Probably you can store some secret ID instead of the actual value and than identify the final record using that ID before flushing the fields.

ivs-cetmix (changes requested):

cetmix_tower_server/models/cx_tower_vault_mixin.py
halbtonjazz: done

cetmix_tower_server/models/cx_tower_vault_mixin.py
halbtonjazz: done

cetmix_tower_server/models/cx_tower_vault_mixin.py
```diff
@@ -0,0 +1,301 @@
...
+        """Create records and set secret values if present"""
+
+        # Step 1: Extract secret fields and generate temporary IDs
+        records_secret_data = []  # [(temp_secret_id, secret_values), ...]
+
+        for vals in vals_list:
+            secret_values = {}
+            for secret_field in self.SECRET_FIELDS:
+                if secret_field in vals and vals.get(secret_field):
+                    secret_values[secret_field] = vals.pop(secret_field)
```
GabbasovDinar: secret_vals = {f: vals.pop(f) for f in self.SECRET_FIELDS if f in vals and vals[f]}

cetmix_tower_server/models/cx_tower_vault_mixin.py
```diff
@@ -0,0 +1,301 @@
...
+    # Single record creation is required because batch creation doesn't allow
+    # matching values from vals_list with the created records when storing
+    # secret fields in the vault. We need the exact record ID to create
+    # corresponding vault entries for each secret field.
+    @api.model_create_multi
+    def create(self, vals_list):
+        """Create records and set secret values if present"""
+
+        # Step 1: Extract secret fields and generate temporary IDs
+        records_secret_data = []  # [(temp_secret_id, secret_values), ...]
```
GabbasovDinar: records_secret_data = []
vault_vals_list = []

cetmix_tower_server/models/cx_tower_vault_mixin.py
```diff
@@ -0,0 +1,301 @@
...
+                # Store secret values in vault with temporary secret_id
+                for field_name, field_value in secret_values.items():
+                    self.env["cx.tower.vault"].sudo().create(
+                        {
+                            "res_model": self._name,
+                            "secret_id": secret_id,  # Temporary ID for linking
+                            "field_name": field_name,
+                            "data": field_value,
+                        }
+                    )
```
GabbasovDinar:
```python
vault_vals_list.append({
                            "res_model": self._name,
                            "secret_id": secret_id,  # Temporary ID for linking
                            "field_name": field_name,
                            "data": field_value,
})
```

cetmix_tower_server/models/cx_tower_vault_mixin.py
```diff
@@ -0,0 +1,301 @@
...
+            else:
+                records_secret_data.append((None, {}))
+
```
GabbasovDinar, line 213:
```python
if vault_vals_list:
    self.env["cx.tower.vault"].sudo().create(vault_vals_list)
```

use create multi instead of create one-by-one

cetmix_tower_server/models/cx_tower_vault_mixin.py
```diff
@@ -0,0 +1,301 @@
...
+                        ]
+                    )
+                )
+
+                # Update to real record ID
+                vault_records.write({"res_id": record.id})
+
+                # Clear secret fields from main table
+                secret_field_list = list(secret_values.keys())
+                record.invalidate_recordset(secret_field_list)
```
GabbasovDinar:
```python
    for record, (secret_id, field_names) in zip(records, records_secret_data):
        if secret_id and field_names:
            vaults = Vault.search([
                ("res_model", "=", self._name),
                ("secret_id", "=", secret_id),
                ("field_name", "in", field_names),
            ])
            vaults.write({"res_id": record.id})
            record.invalidate_recordset(field_names)
```

cetmix_tower_server/models/cx_tower_vault_mixin.py
```diff
@@ -0,0 +1,301 @@
...
+                for record in self:
+                    try:
+                        if record.get_secret_value(secret_field):
+                            record._cache[secret_field] = self.SECRET_VALUE_PLACEHOLDER
+                        else:
+                            record._cache[secret_field] = None
+                    except Exception:  # pylint: disable=except-pass
+                        # skip SpecialValue
+                        # (e.g. for missing record or access right)
+                        pass
```
GabbasovDinar:
```python
show_all = not fields

secret_fields = self.SECRET_FIELDS if show_all else [f for f in self.SECRET_FIELDS if f in fields]

for record in self:
    for secret_field in secret_fields:
        try:
            ....

```

GabbasovDinar (changes requested):

cetmix_tower_server/models/cx_tower_vault_mixin.py
halbtonjazz: added, thank for the suggestion

cetmix_tower_server/models/cx_tower_vault_mixin.py
halbtonjazz: Thanks for the suggestion, cool

cetmix_tower_server/models/cx_tower_vault_mixin.py
halbtonjazz: fixed

cetmix_tower_server/models/cx_tower_vault_mixin.py
halbtonjazz: thanks for optimizing

cetmix_tower_server/models/cx_tower_vault_mixin.py
halbtonjazz: Ok, but I intentionally used enumerate because `zip` gives a pre-commit error requiring `strict=True`. And Ivan mentioned in this PR that strict doesn't work on Python 3.10 which is used in Odoo official docker image for 16.0. But no problem, I've fixed it to use zip and added `strict=True`

cetmix_tower_server/models/cx_tower_vault_mixin.py
halbtonjazz: fixed, thank for the suggestion

GabbasovDinar (approved):

cetmix_tower_server/models/cx_tower_vault_mixin.py
```diff
@@ -0,0 +1,304 @@
...
+            self.env["cx.tower.vault"]
+            .sudo()
+            .search_read(
+                domain,
+                ["res_model", "res_id", "field_name", "data", "secret_id"],
+            )
+        )
+        return vault_records
+
+    def set_secret_values(self, vals):
```
ivs-cetmix: Same here. Such function should be private.

cetmix_tower_server/models/cx_tower_vault_mixin.py
```diff
@@ -0,0 +1,304 @@
...
+            if (
+                vault_record.get("secret_id") == self.secret_id
+                and vault_record.get("field_name") == field_name
+            ):
+                return vault_record.get("data")
+
+        # Return None if no secret value found anywhere
+        return None
+
+    def get_secret_values(self, fields_list=None):
```
ivs-cetmix: Same here

cetmix_tower_server/models/cx_tower_vault_mixin.py
```diff
@@ -0,0 +1,304 @@
...
+                    if record.get_secret_value(secret_field):
+                        record._cache[secret_field] = self.SECRET_VALUE_PLACEHOLDER
+                    else:
+                        record._cache[secret_field] = None
+                except Exception:  # pylint: disable=except-pass
+                    # skip SpecialValue
+                    # (e.g. for missing record or access right)
+                    pass
+
+    def get_secret_value(self, field_name):
```
ivs-cetmix: I think it's better to make this function private. 
@GabbasovDinar your thoughts?

cetmix_tower_server/models/cx_tower_vault_mixin.py
```diff
@@ -0,0 +1,304 @@
...
+
+        for vals in vals_list:
+            # Extract secret fields using dictionary comprehension
+            secret_vals = {
+                f: vals.pop(f) for f in self.SECRET_FIELDS if f in vals and vals[f]
+            }
+
+            if secret_vals:
+                # Generate temporary unique ID for this record
+                secret_id = uuid.uuid4()
```
ivs-cetmix: This is costly method. Besides it's not temporary because it's saved in the db.

cetmix_tower_server/models/cx_tower_server.py
```diff
@@ -124,7 +127,9 @@ class CxTowerServer(models.Model):
         string="SSH Username", required=True, groups="cetmix_tower_server.group_manager"
     )
     ssh_password = fields.Char(
-        string="SSH Password", groups="cetmix_tower_server.group_manager"
+        string="SSH Password",
+        groups="cetmix_tower_server.group_manager",
+        readonly=False,
```
ivs-cetmix: It's `False` by default, no need to add it explicitly once again.

cetmix_tower_server/models/cx_tower_vault.py
```diff
@@ -0,0 +1,43 @@
...
+    res_id = fields.Many2oneReference(
+        string="Resource ID",
+        model_field="res_model",
+        help="ID of the resource that uses this vault",
+    )
+    field_name = fields.Char(
+        required=True,
+        help="Name of the field that contains the secret value",
+    )
+    secret_id = fields.Char(
```
ivs-cetmix: Why do we need to use another field? We already have res_model and res_id for.

ivs-cetmix (changes requested):

cetmix_tower_server/models/cx_tower_vault.py
halbtonjazz: The `secret_id` field is needed to implement your previous suggestion about avoiding single record creation (https://github.com/cetmix/cetmix-tower/pull/347#discussion_r2219137984). During batch creation, we create vault records with `secret_id` before main records exist (no `res_id` yet), then link them afterward using this identifier

cetmix_tower_server/models/cx_tower_vault_mixin.py
halbtonjazz: I looked at how this is implemented in odoo core and did it in a similar way. The previous implementation without resource costs was writing to the main model and then deleting

cetmix_tower_server/models/cx_tower_vault_mixin.py
halbtonjazz: The question was not addressed to me, but this method is intentionally public as it's used extensively outside the class - in `cx_tower_key.py`, `cx_tower_server.py`, `tests`, etс. But if you want to add _, I'll add it, no problem

cetmix_tower_server/models/cx_tower_server.py
halbtonjazz: fixed

cetmix_tower_server/models/cx_tower_vault_mixin.py
ivs-cetmix: It's not used outside the class to much because `cx_tower_key.py` and `cx_tower_server.py` inherit this mixin, so all calls are inside the same class. 
Making it private makes is unavailable for the rpc calls, which can be good from the security point of view.

cetmix_tower_server/models/cx_tower_vault_mixin.py
ivs-cetmix: The main thing is that we don't need this data after it's stored. Which means it must be deleted at the end of the transaction.  So you can populate simple integers because those values will be isolated inside the transaction anyway.

cetmix_tower_server/models/cx_tower_vault.py
ivs-cetmix: Yes, but we don't need to store that data. Not to mention, that you are creating another link to the same record. Check my comment to the mixin file below.

ivs-cetmix (changes requested):

cetmix_tower_server/models/cx_tower_vault_mixin.py
halbtonjazz: fixed

cetmix_tower_server/models/cx_tower_vault_mixin.py
halbtonjazz: fixed

cetmix_tower_server/models/cx_tower_vault_mixin.py
halbtonjazz: I've fixed this as we discussed and replaced UUID-based linking with temporary identifiers

cetmix_tower_server/models/cx_tower_vault.py
halbtonjazz: I've removed the `secret_id` field concept and implemented the temp_identifier approach for vault mixin secret handling

cetmix_tower_server/models/cx_tower_vault_mixin.py
```diff
@@ -0,0 +1,324 @@
...
+        """
+
+        self.ensure_one()
+
+        if field_name not in self.SECRET_FIELDS or (
+            field_name in self._fields and not self._fields[field_name].store
+        ):
+            return None
+
+        query = f"SELECT {field_name} from {self._table} WHERE id=%s"
```
ivs-cetmix: Why are we fetching something from some table(s)?
Is this possible that single value is stored in some specific table while multiple values are stored in another place?

cetmix_tower_server/models/cx_tower_vault_mixin.py
```diff
@@ -0,0 +1,324 @@
...
+
+        # Step 3: Update vault records with real IDs
+        if secret_vals:
+            self._process_secret_values_after_creation(records, secret_vals)
+
+        return records
+
+    def _extract_secret_fields(self, vals_list):
+        """Extract secret fields and replace with temporary identifiers.
+
```
ivs-cetmix, line 287: Please provide a detailed docstring

cetmix_tower_server/models/cx_tower_vault_mixin.py
```diff
@@ -0,0 +1,324 @@
...
+            tuple: (secret_vals dict, modified vals_list)
+        """
+        temp_id_counter = 0
+        secret_vals = {}
+
+        for vals in vals_list:
+            for secret_field in self.SECRET_FIELDS:
+                if secret_field in vals and vals[secret_field]:
+                    temp_id_counter += 1
+                    temp_identifier = f"temp_secret_{temp_id_counter}"
```
ivs-cetmix: `str(temp_id_counter)` will do the same job faster

cetmix_tower_server/models/cx_tower_vault_mixin.py
```diff
@@ -0,0 +1,324 @@
...
+        return secret_vals, vals_list
+
+    def _process_secret_values_after_creation(self, records, secret_vals):
+        """Process secret values after records creation."""
+        records_dict = self._fetch_records_data(records)
+
+        for record_dict in records_dict:
+            self._process_single_record_secrets(record_dict, secret_vals)
+
+    def _fetch_records_data(self, records):
```
ivs-cetmix: This can be put inside _process_secret_values_after_creation as it doesn't contain any specific business logic.

cetmix_tower_server/models/cx_tower_vault_mixin.py
```diff
@@ -0,0 +1,324 @@
...
+
+        # Update database and vault if needed
+        record = self.browse(record_id)
+        if fields_to_clear:
+            record._clear_temp_values(fields_to_clear)
+        if vault_vals:
+            record._set_secret_values(vault_vals)
+
+        # Invalidate cache
+        record.invalidate_recordset(fields_to_clear)
```
ivs-cetmix: Imagine you have created 1000 records.

cetmix_tower_server/models/cx_tower_vault_mixin.py
```diff
@@ -0,0 +1,324 @@
...
+
+    def _fetch_records_data(self, records):
+        """Fetch records data from database."""
+        fields_str = ", ".join(self.SECRET_FIELDS)
+        query = f"SELECT id, {fields_str} FROM {self._table} WHERE id in %s"
+        self.env.cr.execute(query, (tuple(records.ids),))
+        return self.env.cr.dictfetchall()
+
+    def _process_single_record_secrets(self, record_dict, secret_vals):
+        """Process secrets for a single record."""
```
ivs-cetmix: Please provide complete docstring. Here and everywhere. And always.

ivs-cetmix (changes requested):
Please ensure that all(!) functions have complete docstrings before submitting code for review.

cetmix_tower_server/models/cx_tower_vault_mixin.py
halbtonjazz: We cannot store a `secret value` in the vault until a record is created. At the same time, compute and constraints work when a record is created. An example is the constraint `_constraint_ssh_settings`. It will check secret fields such as `ssh_password,` but since there will be no record in the vault at the time of creation, it will fail with an error because it cannot find the field values. For such cases, we write a temporary value for the secret field when creating the model itself, and this case is handled in the `get_secret_value` method. After successfully creating the record, directly in the create method, we clear the secret field from the model (record) itself and write the actual value to the vault. A later call to `get_secret_value` will return the actual value from the vault.

cetmix_tower_server/models/cx_tower_vault_mixin.py
halbtonjazz: I disagree with this statement. f-strings are faster in Python 3.6+ because f-strings are optimized at the bytecode level and work faster than concatenation. In addition, this will reduce code readability and is a step away from modern python standards

cetmix_tower_server/models/cx_tower_vault_mixin.py
halbtonjazz: added

cetmix_tower_server/models/cx_tower_vault_mixin.py
halbtonjazz: fixed

cetmix_tower_server/models/cx_tower_vault_mixin.py
halbtonjazz: fixed

cetmix_tower_server/models/cx_tower_vault_mixin.py
halbtonjazz: added

cetmix_tower_server/models/cx_tower_vault_mixin.py
```diff
@@ -211,7 +211,23 @@ def _set_secret_values(self, vals):
 
     @api.model_create_multi
     def create(self, vals_list):
-        """Create records and set secret values if present"""
+        """Create records and handle secret values securely.
```
ivs-cetmix: This is a core odoo method. It's already documented in the Odoo source code.

cetmix_tower_server/models/cx_tower_vault_mixin.py
ivs-cetmix: Constraint methods are called at the end of transaction. To that moment all the values are already stored.
Besides you store a placeholder, which means that the table is not empty. And the `_read` method returns the placeholder, so even at the ORM level there is a value.
Just below in L85 you are calling the method that does absolutely the same thing: queries the database. 
So why do you need to run another query which will return some irrelevant value?

cetmix_tower_server/models/cx_tower_vault_mixin.py
```diff
@@ -0,0 +1,418 @@
...
+                        record._cache[secret_field] = self.SECRET_VALUE_PLACEHOLDER
+                    else:
+                        record._cache[secret_field] = None
+                except Exception:  # pylint: disable=except-pass
+                    # skip SpecialValue
+                    # (e.g. for missing record or access right)
+                    pass
+
+    def _get_secret_value(self, field_name):
+        """Retrieve the actual secret value for a specific field.
```
ivs-cetmix: Retrieves the actual secret value for a specific field **for a single record**.

cetmix_tower_server/models/cx_tower_key.py
```diff
@@ -443,15 +395,58 @@ def _set_secret_value(self, value=None):
...
+            field_name (str): Name of the secret field to retrieve
+
+        Returns:
+            str or None: The actual secret value, or None if not found
+
+        Note:
+            This method first attempts to retrieve SSH key values from the main
+            table, then falls back to the parent vault implementation for secrets.
+        """
+        value = self._get_ssh_key_from_main_table(field_name)
```
ivs-cetmix: Why are you fetching this value from the main table? 
It must be stored in the vault as any other secret value.

cetmix_tower_server/models/cx_tower_vault_mixin.py
```diff
@@ -0,0 +1,418 @@
...
+                                        Defaults to all SECRET_FIELDS.
+
+        Returns:
+            list: List of vault record dictionaries. Each dictionary contains:
+                  - 'res_model': Model name of the resource
+                  - 'res_id': ID of the main record
+                  - 'field_name': Name of the secret field
+                  - 'data': The actual secret value
+
+                  Example: [{'res_model': 'cx.tower.server', 'res_id': 2,
```
ivs-cetmix: This model is a mixin which is inherited by a model. This function cannot return secret values of any other model. So adding `res_model` doesn't have any practical reason. Same for `res_id`. 
You must return data in a convenient format that will allow to use them later.
Eg in L86 you have to iterate records in a loop. If you have a dict you could have done the same thing much faster.

cetmix_tower_server/models/cx_tower_vault_mixin.py
ivs-cetmix: This is not a concatenation. This is a type conversion. It will be much faster than an f-string.
Regarding readability, this makes no sense here because this is not a variable.
It is a temporary value used inside the function.

ivs-cetmix (changes requested):

cetmix_tower_server/models/cx_tower_vault_mixin.py
halbtonjazz: Constraints are called in the default _create method in models.py, at the end of which the constraints for store fields are checked 
https://github.com/odoo/odoo/blob/c1d3153ed564c0f47a201e5751740857a15d7db2/odoo/models.py#L4270 ,  https://github.com/odoo/odoo/blob/c1d3153ed564c0f47a201e5751740857a15d7db2/odoo/models.py#L1370
Our logic creates a record in the vault only after the model is created (after _create and before checking the constraints themselves), so we need to get this value from somewhere. For this purpose, in the create method, we pass a temporary value for the secret fields to vals to replace the actual secret value. And in order for the constraints to check the completeness of these fields to work, we added a query to the get_secret_value method to check this temporary value in the model itself.
After the entire create method is completed, our real value will be written to the vault, and therefore we will continue to take the value from it

cetmix_tower_server/models/cx_tower_vault_mixin.py
halbtonjazz: Ok, if we're not talking about concatenation, no problem.
Fixed

cetmix_tower_server/models/cx_tower_vault_mixin.py
halbtonjazz: fixed. I added this in response to this message because you asked to check that all(!) functions have complete in this comment https://github.com/cetmix/cetmix-tower/pull/347#pullrequestreview-3092146319
Before implementing the mixin for this method, the documentation was written here https://github.com/cetmix/cetmix-tower/blob/00d228f689eba607f5cfc4c9a4ea372e5c67a5a8/cetmix_tower_server/models/cx_tower_key.py#L76 and here https://github.com/cetmix/cetmix-tower/blob/14ead49b523c6b11beb93271b771891a267c3acd/cetmix_tower_server/models/cx_tower_key_value.py#L111 
I did it by analogy

cetmix_tower_server/models/cx_tower_vault_mixin.py
halbtonjazz: added

cetmix_tower_server/models/cx_tower_key.py
halbtonjazz: This is redundant code from one of the previous implementations, which is now unnecessary

cetmix_tower_server/models/cx_tower_vault_mixin.py
halbtonjazz: I agree about res_model. res_id should be kept because this method is called on more than one record. As for the format, it is implemented according to the Function details from the task description

cetmix_tower_server/models/cx_tower_vault_mixin.py
ivs-cetmix: Ok, but we put temporary values instead of the secret ones, so there will be no difference from the constraint point of view.
And if we are talking about `_constraint_ssh_settings`, this constraint is defined at the model level, not at the mixin one. So it should be addressed at the model level either.

cetmix_tower_server/models/cx_tower_vault_mixin.py
```diff
@@ -0,0 +1,417 @@
...
+
+        Note:
+            Called automatically during create() process. Should not be used directly.
+        """
+        fields_str = ", ".join(self.SECRET_FIELDS)
+        query = f"SELECT id, {fields_str} FROM {self._table} WHERE id in %s"
+        self.env.cr.execute(query, (tuple(records.ids),))
+        records_dict = self.env.cr.dictfetchall()
+
+        fields_to_invalidate = set()
```
ivs-cetmix: You can simply invalidate all secret fields. This will not affect performance and will make the code simpler.

cetmix_tower_server/models/cx_tower_vault_mixin.py
```diff
@@ -0,0 +1,417 @@
...
+        # Update database and vault if needed
+        record = self.browse(record_id)
+        if fields_to_clear:
+            record._clear_temp_values(fields_to_clear)
+        if vault_vals:
+            record._set_secret_values(vault_vals)
+
+        return fields_to_clear
+
+    def _clear_temp_values(self, fields_to_clear):
```
ivs-cetmix: Same here. What if we have 1000 records? It's much easier and faster to clear all secret fields for the entire recordset without checking them and clearing selectively.

cetmix_tower_server/models/cx_tower_vault_mixin.py
```diff
@@ -0,0 +1,417 @@
...
+        # Extract secret fields
+        secret_values = {}
+        for secret_field in self.SECRET_FIELDS:
+            if secret_field in vals:
+                secret_values[secret_field] = vals.pop(secret_field)
+
+        if secret_values:
+            self._set_secret_values(secret_values)
+            # Invalidate cache for all secret fields
+            for field_name in secret_values.keys():
```
ivs-cetmix: 20 fields = 20 invalidations of the same recordset

cetmix_tower_server/models/cx_tower_vault_mixin.py
```diff
@@ -0,0 +1,417 @@
...
+        ]
+        vault_records = (
+            self.env["cx.tower.vault"]
+            .sudo()
+            .search_read(
+                domain,
+                ["res_id", "field_name", "data"],
+            )
+        )
+        return vault_records
```
ivs-cetmix: The original task stated you should return the result in the `read` function format. This is not a `read` function format because it contains `res_id` instead of `id`.
Anyway to make this function more usable we should implement a more convenient format, eg a dict. Same way as it is done [here](https://github.com/cetmix/cetmix-tower/blob/00d228f689eba607f5cfc4c9a4ea372e5c67a5a8/cetmix_tower_server/models/cx_tower_variable_mixin.py#L36).

ivs-cetmix (changes requested):

cetmix_tower_server/models/cx_tower_vault_mixin.py
halbtonjazz: fixed

cetmix_tower_server/models/cx_tower_vault_mixin.py
halbtonjazz: fixed

cetmix_tower_server/models/cx_tower_vault_mixin.py
halbtonjazz: fixed

cetmix_tower_server/models/cx_tower_vault_mixin.py
halbtonjazz: fixed

cetmix_tower_server/models/cx_tower_vault_mixin.py
```diff
@@ -0,0 +1,385 @@
...
+                            "field_name": field,
+                            "data": value,
+                        }
+                    )
+
+        # Batch operations
+        for value, records in vals_to_update_records.items():
+            records.sudo().write({"data": value})
+
+        self.env["cx.tower.vault"].sudo().create(records_to_create)
```
ivs-cetmix:
```python
if records_to_create:
.....
```

cetmix_tower_server/models/cx_tower_vault_mixin.py
```diff
@@ -0,0 +1,385 @@
...
+        records_to_unlink.sudo().unlink()
```
ivs-cetmix:
```python
if records_to_unlink:
....
```

cetmix_tower_server/models/cx_tower_vault_mixin.py
```diff
@@ -0,0 +1,385 @@
...
+        )
+
+        # Collect secret values and fields to clear
+        for field_name, temp_identifier in field_temp_id_pairs:
+            secret_value = secret_vals.get(temp_identifier)
+            if secret_value:
+                vault_vals[field_name] = secret_value
+
+        # Update database and vault if needed
+        record = self.browse(record_id)
```
ivs-cetmix: We need the record only if we have the vault_vals

cetmix_tower_server/models/cx_tower_vault_mixin.py
```diff
@@ -0,0 +1,385 @@
...
+            self._set_secret_values(secret_values)
+            # Invalidate cache for all secret fields
+            self.invalidate_recordset(self.SECRET_FIELDS)
+
+        return super().write(vals)
+
+    def unlink(self):
+        """Override unlink to delete vault records.
+
+        Automatically removes all associated vault records before deleting
```
ivs-cetmix: What if there is an error in the unlink method and transaction is rolled back?

cetmix_tower_server/models/cx_tower_key.py
```diff
@@ -443,15 +395,16 @@ def _set_secret_value(self, value=None):
...
+                create_vals = {"key_id": self.id}
+                create_vals.update(vals)
+                self.value_ids.create(create_vals)
 
         elif self.key_type == "k":
-            self.with_context(write_secret_value=True).write({"secret_value": value})
-            self.invalidate_recordset(["secret_value"])
+            secret_fields_list = list(vals.keys())
+            set_clause = ", ".join(f"{field} = %s" for field in secret_fields_list)
+            query = f"UPDATE {self._table} SET {set_clause} WHERE id = %s"
```
ivs-cetmix: Why are we using an SQL query instead of generic functions? Actually why do we need to do anything for the "k" type key? Wouldn't it be handled by the mixin?

cetmix_tower_server/models/cx_tower_server.py
```diff
@@ -1950,3 +1905,21 @@ def _get_notification_action(
...
+
+    @api.model_create_multi
+    def create(self, vals_list):
+        """Override create to validate SSH password before record creation."""
+        # Validate SSH password before creating records
+        if not self._context.get("skip_ssh_settings_check"):
+            validation_errors = []
+            for vals in vals_list:
+                if vals.get("ssh_auth_mode") == "p" and not vals.get("ssh_password"):
+                    server_name = vals.get("name", "Unknown Server")
```
ivs-cetmix: "name" is a mandatory field

ivs-cetmix (changes requested):

cetmix_tower_server/models/cx_tower_key.py
```diff
@@ -86,25 +89,10 @@ def create(self, vals_list):
             secret_value = vals.pop("secret_value", None)
             new_key = super().create(vals)
             if secret_value:
-                new_key._set_secret_value(secret_value)
+                new_key._set_secret_values({"secret_value": secret_value})
```
ivs-cetmix: This is not needed any longer because mixin will do the job for ssh keys.
So we need to split the keys by type - ssh or secret.
For ssh keys we can simply call super.
For secrets we need to create them one by one as before while adjusting this process to the new mixin flow.

ivs-cetmix (changes requested):

cetmix_tower_server/models/cx_tower_key.py
halbtonjazz: I analyzed this fragment and realized I made an error. I didn't need to override the create method in this model. Because the mixin's create method already handles secret field values and calls the overridden _set_secret_values from cx.tower.key, which stores the secret value depending on the type - secrets or SSH keys

ivs-cetmix (changes requested):
Code LGTM, however you must follow the OCA coding guideline when in comes to adding functions if Python files.

ivs-cetmix (approved):
Intermediate code review LGTM

CetmixGitDrone:

GabbasovDinar (approved):

PR #465, opened 2026-01-29
[16.0][IMP] cetmix_tower_server Jets

ivs-cetmix:
Jets are here! 🚀 
Opening a new PR because https://github.com/cetmix/cetmix-tower/pull/375 became really huuuuuge, so it's not possible to open it in the browser any longer 😕 

## What are Jets?

Jets are designed to atomise the current Cetmix Tower Server entity making both deployment and management of software more flexible and convenient.
Jet is a sort of a component that has its own lifecycle and dependencies. Jet's lifecycle is defined by states and transition between those states is triggered by Actions.
Action takes a Jet from its initial state and moves it to the destination state via the transitional state. In case if transition fails Jet rolls back to the initial state or to the on-error state if such is defined. The transition action is handled by a Flight Plan which can be assigned to the Action. If no Flight Plan is assigned to an Action the related transition is performed instantly.

Jets are created from the Jet Templates which define both dependencies and lifecycle of Jets created from them.
To be able to launch a Jet on a Server the corresponding Jet Template must be first installed on the Server.
For such installations Jet Template Installations are used which install the selected Jet Template together with all the Jet Templates the selected Jet Template depends on.
Jet Template can have a Flight Plan which is executed when this Jet Template is being installed on a Server. If such Flight Plan fails, the entire Jet Template Installation is marked as failed. However if some of the dependencies were installed successfully, those ones remain installed and will not be reinstalled again during the following installations.
If a Jet Template doesn't have an installation Flight Plan defined, such Jet Template will be installed instantly.

Task 4700

## Summary by CodeRabbit

* **New Features**
  * Introduces a complete "Jet" system: templates, states, actions, dependencies, lifecycle/state machine, install/uninstall/clone flows, waypoints, requests, wizards, UI views, command/flight-plan integration, logs, files and rich demo data.

* **Chores**
  * Adds backend refresh support, migration cleanup, URL validation, expanded security/access rules and installation orchestration.

* **Tests**
  * Large suite of new tests covering jets, templates, installs, waypoints, actions and access control.

* **Documentation**
  * Diagrams illustrating Jet flows, requests and state transitions.


cetmix_tower_server/models/cx_tower_command_log.py
```diff
@@ -122,6 +143,17 @@ def _compute_name(self):
...
+    @api.depends("plan_log_id")
+    def _compute_jet_id(self):
+        for command_log in self:
+            if command_log.plan_log_id:
+                command_log.update(
+                    {
+                        "jet_id": command_log.plan_log_id.jet_id,
+                        "jet_template_id": command_log.plan_log_id.jet_template_id,
+                    }
+                )
```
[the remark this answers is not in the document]
ivs-cetmix: If a command is run directly, not from a flight plan, then these values will be written directly.

cetmix_tower_server/models/cx_tower_jet_request.py
```diff
@@ -0,0 +1,254 @@
...
+        )
+        if transitioning_jets:
+            _logger.info(
+                "Jet %s is transitioning to the target state, "
+                "waiting for it to finish",
+                transitioning_jets[0].name,
+            )
+            # Trigger the jet to bring itself to the required state
+            request.jet_id = transitioning_jets[0]
+            return request
```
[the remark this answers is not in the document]
ivs-cetmix: A Jet cannot server a request while transitioning between states. Once the transition is finished the `_finalize_transition()` will trigger the `_on_is_available()` method that will pick up the request.

cetmix_tower_server/models/cx_tower_jet_state.py
```diff
@@ -0,0 +1,91 @@
...
+        # his access level is considered as user.
+        # NB: record access is already checked above.
+        if user_access_level == "2" and self.env.user not in jet.manager_ids:
+            user_access_level = "1"
+
+        # Check if user access level is equal or greater
+        if self.access_level > user_access_level:
+            raise AccessError(
+                _("You are not allowed to set the '%(state)s' state!", state=self.name)
+            )
```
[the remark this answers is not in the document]
ivs-cetmix: Transitions are done using sudo(). User will always have an access level as it's defined by the group membership. And without being a group member user cannot even open the Tower.

cetmix_tower_server/models/cx_tower_jet_template_dependency.py
```diff
@@ -0,0 +1,165 @@
...
+
+            if from_id not in graph:
+                graph[from_id] = set()
+
+            graph[from_id].add(to_id)
+
+            # Ensure the to_id is in the graph even if it doesn't require anything
+            if to_id not in graph:
+                graph[to_id] = set()
+
```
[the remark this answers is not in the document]
ivs-cetmix: This method is called from behind of the root user when used in the compute method of the storable computed field.

cetmix_tower_server/models/cx_tower_jet_template_install.py
```diff
@@ -0,0 +1,481 @@
...
+        self.env.user.reload_views(
+            model="cx.tower.server", view_types=["form"], rec_ids=[self.server_id.id]
+        )
+        self.env.user.reload_views(
+            model="cx.tower.jet.template",
+            view_types=["form"],
+            rec_ids=[self.jet_template_id.id],
+        )
+
+        # Check if notifications are enabled
```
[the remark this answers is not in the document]
ivs-cetmix: run_flight_plan() will return a error if exception happens during its execution.

cetmix_tower_server/models/cx_tower_jet_template.py
```diff
@@ -0,0 +1,1399 @@
...
+                )
+            )
+        templates_with_installed_servers = self.filtered(lambda t: t.server_ids)
+        if templates_with_installed_servers:
+            raise ValidationError(
+                _(
+                    "Following templates cannot be deleted "
+                    "as they are installed on servers: %s",
+                    ",".join(templates_with_installed_servers.mapped("display_name")),
+                )
```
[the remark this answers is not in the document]
ivs-cetmix: Your suggestion is good, however this approach will work as well

cetmix_tower_server/models/cx_tower_jet_waypoint_template.py
```diff
@@ -0,0 +1,55 @@
...
+        """
+        Available access levels
+
+        Returns:
+            List of tuples: available options.
+        """
+        return [
+            ("2", "Manager"),
+            ("3", "Root"),
+        ]
```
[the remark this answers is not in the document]
ivs-cetmix: These methods are inherited from mixin, so we need to override them in each model separatelly.

cetmix_tower_server/models/cx_tower_jet_waypoint.py
```diff
@@ -0,0 +1,541 @@
...
+                    and waypoint.state != "draft"
+                ):
+                    raise ValidationError(
+                        _(
+                            "Cannot change waypoint type for %(waypoint)s "
+                            "because it is not in the draft state",
+                            waypoint=waypoint.name,
+                        )
+                    )
+        return super().write(vals)
```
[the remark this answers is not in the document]
ivs-cetmix: This is fine because it means that the new state will be draft.

cetmix_tower_server/models/cx_tower_jet.py
```diff
@@ -0,0 +1,1610 @@
...
+
+        # Check if the jet is deletable
+        not_deletable_jets = self.filtered(lambda j: not j.deletable)
+        if not_deletable_jets:
+            raise ValidationError(
+                _(
+                    "Following jets cannot be deleted as they are not deletable: %s",
+                    not_deletable_jets.mapped("display_name"),
+                )
+            )
```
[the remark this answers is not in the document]
ivs-cetmix: Thank you for noticing, this will work either. You can check here: https://www.odoo.com/documentation/18.0/developer/howtos/translations.html#variables

cetmix_tower_server/models/cx_tower_jet.py
```diff
@@ -0,0 +1,1610 @@
...
+        Check if the jet is available to accept new links from other jets.
+
+        Returns:
+            bool: True if the jet is available to accept new links from other jets,
+                False otherwise
+        """
+        self.ensure_one()
+        # TODO: Implement the logic to check if the jet is available
+        # to accept new links from other jets
+        return True
```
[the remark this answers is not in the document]
ivs-cetmix: We will add them later.

cetmix_tower_server/models/cx_tower_scheduled_task.py
```diff
@@ -31,6 +36,20 @@ class CxTowerScheduledTask(models.Model):
...
+        "jet_id",
+        string="Jets",
+    )
+    jet_template_ids = fields.Many2many(
+        string="Jet Templates",
+        comodel_name="cx.tower.jet.template",
+        relation="cx_tower_jet_template_scheduled_task_rel",
+        column1="scheduled_task_id",
+        column2="jet_template_id",
+    )
```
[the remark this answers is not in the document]
ivs-cetmix: This is because this field is used only to link tasks with jet templates.

cetmix_tower_server/models/cx_tower_server_log.py
```diff
@@ -64,6 +70,25 @@ class CxTowerServerLog(models.Model):
...
+    jet_id = fields.Many2one(
+        "cx.tower.jet",
+        ondelete="cascade",
+    )
+
+    @api.depends("jet_id")
+    def _compute_server_id(self):
+        for record in self:
+            if not record.server_id and record.jet_id:
+                record.server_id = record.jet_id.server_id.id
```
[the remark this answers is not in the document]
ivs-cetmix: In other cases server_id is set explicitly.

cetmix_tower_server/models/cx_tower_variable.py
```diff
@@ -362,7 +370,515 @@ def _validate_value(self, value_char=None):
...
+        if server:
+            return server_value_char or global_value_char
+
+        # 2.5. Plan Line Action
+        if plan_line_action:
+            return plan_line_action_value_char
+
+        # 2.6. Global
+        return global_value_char
+
```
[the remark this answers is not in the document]
ivs-cetmix: Plan line action values are not used in the evaluation context

cetmix_tower_server/models/res_users.py
```diff
@@ -0,0 +1,29 @@
...
+            - "3": Root
+        """
+
+        if self.env.user.has_group("cetmix_tower_server.group_root"):
+            return "3"
+        elif self.env.user.has_group("cetmix_tower_server.group_manager"):
+            return "2"
+        elif self.env.user.has_group("cetmix_tower_server.group_user"):
+            return "1"
+        else:
```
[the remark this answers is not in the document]
ivs-cetmix: Yes, this is we want: check the current user, not the one in the record.

cetmix_tower_server/models/tools.py
```diff
@@ -33,3 +34,42 @@ def get_section():
...
+    if not host:
+        return False
+
+    if host in ("localhost", "::1"):
+        return True
+
+    if "." in host or ":" in host:
+        return True
+
+    return False
```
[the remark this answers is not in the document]
ivs-cetmix: This is not needed

cetmix_tower_server/security/cx_tower_jet_security.xml
```diff
@@ -0,0 +1,93 @@
...
+                "|",
+                    ("server_id.user_ids", "in", [user.id]),
+                    ("server_id.manager_ids", "in", [user.id])
+            ]
+        </field>
+        <field name="perm_read" eval="0" />
+        <field name="perm_write" eval="1" />
+        <field name="perm_create" eval="1" />
+        <field name="perm_unlink" eval="0" />
+    </record>
```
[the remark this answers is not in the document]
ivs-cetmix: Let's leave it as is

cetmix_tower_server/security/cx_tower_jet_template_install_line_security.xml
```diff
@@ -0,0 +1,34 @@
...
+        <field name="domain_force">
+            ["&amp;",
+                "|",
+                    ("server_id.user_ids", "in", [user.id]),
+                    ("server_id.manager_ids", "in", [user.id]),
+                "|",
+                    ("jet_template_id.access_level", "&lt;=", "2"),
+                    ("jet_template_id.user_ids", "in", [user.id])
+            ]
+        </field>
```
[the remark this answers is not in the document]
ivs-cetmix: Yes, because when installing a template on server user must have access to both template and server

cetmix_tower_server/security/cx_tower_jet_template_install_security.xml
```diff
@@ -0,0 +1,34 @@
...
+        <field name="perm_unlink" eval="0" />
+    </record>
+
+    <!-- Root Full Access -->
+    <record id="rule_cx_tower_jet_template_install_root_full" model="ir.rule">
+        <field name="name">Jet Template Install: Root Full Access</field>
+        <field name="model_id" ref="model_cx_tower_jet_template_install" />
+        <field name="groups" eval="[(4, ref('cetmix_tower_server.group_root'))]" />
+        <field name="domain_force">[(1, '=', 1)]</field>
+    </record>
```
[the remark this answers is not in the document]
ivs-cetmix: Let's leave it as is

cetmix_tower_server/tests/test_jet_template_install.py
```diff
@@ -0,0 +1,1754 @@
...
+                line.jet_template_id,
+                expected_template,
+                f"Line {i} should be {expected_template.name}",
+            )
+            self.assertEqual(
+                line.order,
+                expected_order_num,
+                f"Line {i} should have order {expected_order_num}",
+            )
+
```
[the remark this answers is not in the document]
ivs-cetmix: Let's leave it as is for now

cetmix_tower_server/tests/test_jet.py
```diff
@@ -0,0 +1,1392 @@
...
+
+        # Manager not in manager_ids should have effective access level "1"
+        effective_level = self.jet_test.with_user(
+            self.manager
+        )._get_user_effective_access_level()
+        self.assertEqual(
+            effective_level,
+            "1",
+            "Manager not in manager_ids should have effective access level 1",
+        )
```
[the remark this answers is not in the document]
ivs-cetmix: Each test runs in its own transaction that rolls back, so other tests don't affect this one.
 setUpClass doesn't set server_test_1.manager_ids, it's created without manager_ids.

cetmix_tower_server/models/cx_tower_variable.py
```diff
@@ -362,7 +370,515 @@ def _validate_value(self, value_char=None):
...
+                # Render variables used in value
+                values_for_value = self._get_variable_values_by_references(
+                    value_vars,
+                    apply_modifiers=True,
+                    server=server,
+                    jet_template=jet_template,
+                    jet=jet,
+                    _depth=_depth,
+                )
```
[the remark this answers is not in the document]
ivs-cetmix: Server template and plan line actions are not used in the recursive rendering.

cetmix_tower_server/models/cx_tower_command.py
```diff
@@ -35,10 +36,50 @@
...
+        "quote_from_bytes",
+        "unquote",
+        "unquote_plus",
+        "unquote_to_bytes",
+        "splitquery",
+        "splittag",
+        "splituser",
+        "splitvalue",
+    ],
+)
```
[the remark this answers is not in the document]
ivs-cetmix: Noted, let's remove them completely

cetmix_tower_server/models/cx_tower_server.py
```diff
@@ -1148,12 +1323,17 @@ def _command_runner(
                 rendered_command_code,
                 **kwargs,
             )
+        elif command.action == "jet_action":
+            response = self._command_runner_jet_action(
+                log_record,
+                **kwargs,
+            )
```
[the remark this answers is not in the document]
ivs-cetmix: Check again if this is still a valid issue

cetmix_tower_server/models/cx_tower_variable.py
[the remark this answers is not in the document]
ivs-cetmix: server_template/plan_line_action are not used in nested templates

cetmix_tower_server/models/cx_tower_jet_dependency.py
```diff
@@ -0,0 +1,65 @@
...
+            if record.jet_id == record.jet_depends_on_id:
+                raise ValidationError(_("A jet cannot depend on itself!"))
+            # Ensure jet that we depend on has the templated
+            # from the template dependency
+            if (
+                record.jet_depends_on_id
+                and record.jet_template_dependency_id
+                and record.jet_depends_on_id.jet_template_id
+                != record.jet_template_dependency_id.template_required_id
+            ):
```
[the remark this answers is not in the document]
ivs-cetmix: Check again if this is still a valid issue

cetmix_tower_server/models/cx_tower_jet_template_install.py
```diff
@@ -0,0 +1,487 @@
...
+            title=template.name,
+            sticky=False,  # explicitly set to False to avoid blocking the user's screen
+            action=action,
+        )
+
+        # Launch the installation
+        install_record._process_install()
+
+        # Return the installation record
+        return install_record
```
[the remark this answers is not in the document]
ivs-cetmix: Both functions have "@api.model" decorator, so they don't contain recordsets

cetmix_tower_server/models/cx_tower_jet_dependency.py
```diff
@@ -0,0 +1,63 @@
...
+        help="Jet this Jet depends on.",
+        ondelete="cascade",
+    )
+
+    _sql_constraints = [
+        (
+            "unique_jet_dependency",
+            "UNIQUE(jet_id, jet_depends_on_id)",
+            "This dependency already exists!",
+        )
```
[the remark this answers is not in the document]
ivs-cetmix: This is ok, dependent jet can be added later

cetmix_tower_server/models/cx_tower_jet.py
```diff
@@ -0,0 +1,1628 @@
...
```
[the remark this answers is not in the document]
ivs-cetmix: Let's leave as if for now

cetmix_tower_server/models/cx_tower_server.py
```diff
@@ -956,54 +1119,55 @@ def _render_command(self, command, path=None, custom_variable_values=None):
...
+                .sudo()
+                ._get_variable_values_by_references(
+                    variable_references, server=self, jet_template=jet_template, jet=jet
+                )
             )
 
+            # Apply custom variable values only if user has write access to the server
+            has_write_access = self._have_access_to_server("write")
+            if custom_variable_values and has_write_access:
+                variable_values.update(custom_variable_values)
```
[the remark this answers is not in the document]
ivs-cetmix: Let's leave as is of now

cetmix_tower_server/models/cx_tower_jet.py
```diff
@@ -0,0 +1,1628 @@
...
+
+        Returns:
+            cx.tower.jet(): Recordset of jets
+        """
+        self.ensure_one()
+
+        jet_template = self.jet_template_id.get_by_reference(jet_template_reference)
+        if jet_template:
+            return self._get_dependent_jets_by_template(jet_template)
+        return False
```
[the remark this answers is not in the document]
ivs-cetmix: Thank you, let's leave as is for now. Because the result is always checked using `if jets: ...`

cetmix_tower_server/models/cx_tower_jet.py
```diff
@@ -0,0 +1,1628 @@
...
+                    request._finalize(failed=False)
+
+            # Pick the first request that requests a different state
+            remaining_requests = implicit_requests - same_state_requests
+            if remaining_requests:
+                remaining_request = remaining_requests[0]
+                # Set current jet as the target jet for the request
+                remaining_request.write({"jet_id": self.id})  # pylint: disable=no-member
+                self._bring_to_state(remaining_request.state_requested_id)
+
```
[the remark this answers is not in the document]
ivs-cetmix: Let's leave as is for now

cetmix_tower_server/models/cx_tower_server.py
```diff
@@ -521,6 +584,56 @@ def action_open_files(self):
...
+        # Open the wizard to install the template on the selected servers
+        return {
+            "type": "ir.actions.act_window",
+            "name": _("Install Jet Template"),
+            "res_model": "cx.tower.jet.template.install.wiz",
+            "view_mode": "form",
+            "target": "new",
+            "context": {
+                "default_server_ids": self.ids,  # pylint: disable=no-member
+            },
```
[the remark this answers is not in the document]
ivs-cetmix: It's done on purpose

cetmix_tower_server/models/cx_tower_server.py
```diff
@@ -1461,6 +1644,95 @@ def _command_runner_python_code(
...
+                template=requested_jet_template.name,
+            )
+
+            log_record.finish(
+                status=status,
+                response=response,
+                error=error,
+            )
+        # Return result
+        return {"status": status, "response": response, "error": error}
```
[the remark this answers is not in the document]
ivs-cetmix: Is this still a valid issue?

cetmix_tower_server/demo/demo_data.xml
```diff
@@ -565,6 +566,77 @@ else:
         />
     </record>
 
+    <!-- Installation Commands -->
+    <record id="command_odoo_install" model="cx.tower.command">
+        <field name="name">Install Odoo</field>
+        <field name="action">python_code</field>
+        <field name="code">#ok</field>
```
SirPyTech, line 573: **question**: I understand this is demo data, but other demo commands are actually doing stuff (`mkdir` and `apt-get update` for instance), is it correct that these new ones only contain comments in code?

cetmix_tower_server/models/cetmix_tower.py
```diff
@@ -2,34 +2,29 @@
...
     _name = "cetmix.tower"
     _description = "Cetmix Tower Odoo Automation"
 
     @api.model
     def server_create_from_template(self, template_reference, server_name, **kwargs):
-        """Shortcut for the same method of the 'cx.tower.server.template' model.
-
-        Important! Add dedicated tests for this function if modified later.
+        """
+        THIS METHOD IS DEPRECATED. USE THE 'cx.tower.server.template' MODEL DIRECTLY.
```
SirPyTech, line 27: **note**: Deprecations in comments are usually ignored until it's too late, a warning log could be more effective and would show in the CI.

cetmix_tower_server/models/cx_tower_command.py
```diff
@@ -185,6 +225,28 @@ class CxTowerCommand(models.Model):
             "after running the command.\n"
         ),
     )
+    # -- Jets
+    jet_template_id = fields.Many2one(
+        comodel_name="cx.tower.jet.template",
+        help="Action will be triggered for all dependent jets" " of this template",
```
SirPyTech, line 231: **question**: This command will run on all jets generated by this template?
I don't see it in `cx.tower.server._run_python_code`, or maybe I misunderstood something int what we are trying to achieve.

More in general about jet templates usage: I understand they have references to many objects (commands, files, ...) because the jets they generate will need similar links, but there are some runtime behaviors (like executing a command on a template or getting its log) that I was not expecting in a template, could you please clarify their scope?

cetmix_tower_server/models/cx_tower_jet_dependency.py
```diff
@@ -0,0 +1,63 @@
...
+            if record.jet_id == record.jet_depends_on_id:
+                raise ValidationError(_("A jet cannot depend on itself!"))
+            # Ensure jet that we depend on has the template
+            # from the template dependency
+            if (
+                record.jet_depends_on_id
+                and record.jet_template_dependency_id
+                and record.jet_depends_on_id.jet_template_id
+                != record.jet_template_dependency_id.template_required_id
+            ):
```
SirPyTech, line 60: **suggestion**: if I understand the comment correctly:
```suggestion
            # Ensure jet that we depend on has the template
            # from the template dependency
            if (
                record.jet_depends_on_id
                and record.jet_template_dependency_id
                and record.jet_depends_on_id.jet_template_id
                != record.jet_template_dependency_id
            ):
```

cetmix_tower_server/models/cx_tower_jet.py
```diff
@@ -0,0 +1,1617 @@
...
+        """Compute the domain of the jet template"""
+        for jet in self:
+            jet.jet_template_domain = (
+                [("server_ids", "in", [jet.server_id.id])] if jet.server_id else []
+            )
+            jet.server_allowed_domain = (
+                [("id", "in", jet.jet_template_id.server_ids.ids)]
+                if jet.jet_template_id
+                else []
+            )
```
SirPyTech, line 268: **suggestion**:
```suggestion
    @api.depends("server_id", "jet_template_id")
    def _compute_jet_template_domain(self):
        """Compute the domain of the jet template"""
        for jet in self:
            jet.jet_template_domain = (
                [("server_ids", "in", [jet.server_id.id])] if jet.server_id else []
            )
            jet.server_allowed_domain = (
                [("id", "in", jet.jet_template_id.server_ids.ids)]
                if jet.jet_template_id.server_ids
                else []
            )
```
Or maybe create a different compute method since the two domains have nothing in common.

cetmix_tower_server/models/res_users.py
```diff
@@ -0,0 +1,30 @@
...
+            return False
```
SirPyTech, line 34: **suggestion**: This is a risky abstraction in my opinion, I'd stick with using groups where needed: what if a middle group appears? Or if they are not completely ordered?
If you want to keep it, at least use constants like `ROOT = "3"` or an `Enum` instead of the 3, 2, 1 strings directly so that usages can be easily tracked down.

cetmix_tower_server/models/cx_tower_jet.py
```diff
@@ -0,0 +1,1617 @@
...
+            cx.tower.jet.waypoint or False: The created waypoint or False if failed.
+        """
+        self.ensure_one()
+
+        # Check if the jet is busy
+        if self._is_busy() and not ignore_busy:
+            _logger.error(
+                "Cannot create waypoint for jet %s because it is busy", self.name
+            )
+            return False
```
SirPyTech: **suggestion**: The user should probably be notified that something has failed, why not a `UserError`?

cetmix_tower_server/models/cx_tower_jet_waypoint.py
```diff
@@ -0,0 +1,667 @@
...
+    _inherit = [
+        "cx.tower.reference.mixin",
+        "cx.tower.access.mixin",
+        "cx.tower.metadata.mixin",
+    ]
+    _order = "create_date desc"
+
+    name = fields.Char(required=True)
+    access_level = fields.Selection(
+        compute="_compute_access_level", readonly=False, store=True
```
SirPyTech: **issue**: I can't find this `_compute_access_level` method here or in any of the inherited models, where is it?

SirPyTech (commented):
Jets are going to be awesome!
I'm looking forward to the talk in the OCA Days about them 🤗

I only had time to review part of the code, please see the comments.

cetmix_tower_server/demo/demo_data.xml
ivs-cetmix: This command runs Python code. By default [Python commands](https://cetmix.com/tower/documentation/reference/execution/command/#python-code-commands) return `result = {"exit_code": 0, "message": Null}`. So an empty command with just one commented line will always finish with the `success` result.

cetmix_tower_server/models/cetmix_tower.py
ivs-cetmix: Thank you! Will add this.

cetmix_tower_server/models/cx_tower_command.py
ivs-cetmix: Yes, but only if these jets are somehow linked to the Jet that is triggering this command.
Example: when a Postgres jet stops we need to stop all the Odoo jets that are depend on it.
In this case the template here will be "Odoo". 
Tower will check all the jets with the template "Odoo" in the current Postgres jet dependencies and trigger the "Stop" action for them.

cetmix_tower_server/models/res_users.py
ivs-cetmix: Sounds reasonable!

cetmix_tower_server/models/cx_tower_jet.py
ivs-cetmix: This method is mostly used in automation, so the idea is that the other method that calls it should check the result and behave based on it, e.g. raise an exception.

cetmix_tower_server/models/cx_tower_jet_waypoint.py
ivs-cetmix: My bad, forgot to add the method 🤦

cetmix_tower_server/models/cx_tower_jet_waypoint.py
```diff
@@ -0,0 +1,541 @@
...
+    def _compute_display_name(self):
+        """
+        Compute the display name of the waypoint
+        """
+        for waypoint in self:
+            timestamp = fields.Datetime.context_timestamp(
+                waypoint, waypoint.create_date
+            )
+            formatted_date = timestamp.strftime("%Y-%m-%d %H:%M:%S")
+            waypoint.display_name = f"{waypoint.name} ({formatted_date})"
```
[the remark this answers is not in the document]
ivs-cetmix: this is a deafault field that is always set.

cetmix_tower_server/models/cx_tower_variable.py
```diff
@@ -362,7 +370,518 @@ def _validate_value(self, value_char=None):
...
+                "type": jet.waypoint_id.waypoint_template_id.reference
+                if jet.waypoint_id
+                else False,
+            }
+            # Add each metadata key-value pair to the waypoint data
+            metadata = jet.waypoint_id.metadata if jet.waypoint_id else False
+            if metadata:
+                for key, value in metadata.items():
+                    waypoint_data[key] = value
+            values.update({"waypoint": waypoint_data})
```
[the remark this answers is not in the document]
ivs-cetmix: This is done on purpose.

cetmix_tower_server/models/cx_tower_jet_dependency.py
ivs-cetmix: Initial code is should be correct, because jet_template_dependency refers the `cx.tower.jet.template.dependency` Class, not the `cx.tower.jet.template`.

cetmix_tower_server/models/res_users.py
ivs-cetmix: Checked this once again. Although I share your concerns there is the following reason behind this:
We are deliberately using "3","2", "1" pattern so we can use [comparison instructions](https://github.com/cetmix/cetmix-tower/blob/99aa8c621c3903f2f4d98a5a7d8766f87c521a0e/cetmix_tower_server/security/cx_tower_jet_template_security.xml#L31) in the access rules.
So this patter is not a subject for some future changes, unless there will be some apocalyptic refactoring 😄

cetmix_tower_server/demo/demo_data.xml
SirPyTech: I see, thanks for the clarification; what I meant is that other commands really do what they say, like "Create directory" actually creates a directory using `mkdir`:
```xml
    <record id="command_create_dir" model="cx.tower.command">
        <field name="name">Create directory</field>
        <field name="action">ssh_command</field>
        <field
            name="server_ids"
            eval="[(6, 0, [ref('cetmix_tower_server.server_demo_1')])]"
        />
        <field name="path">/home/{{ tower.server.username }}</field>
        <field name="code">mkdir -p {{ demo_dir }}</field>
        <field
            name="tag_ids"
            eval="[(6, 0, [ref('cetmix_tower_server.tag_production')])]"
        />
        <field name="note">Create a directory on the host system</field>
        <field name="access_level">1</field>
    </record>
```
while this "Install Odoo" does nothing, is this on purpose?

cetmix_tower_server/models/cetmix_tower.py
SirPyTech: It's not there yet.

cetmix_tower_server/models/cx_tower_command.py
SirPyTech: Thanks, the example helps!
The "Stop" action in the Postgres jet triggers the "Stop" action for all the Odoo jets that have been deployed as part of the same template, this makes sense :+1:

So I probably misunderstood some parts of the code: 
- about commands: start/stop are never executed on templates directly.
- about logs: the template field is only used to aggregate the logs of all the jets it has generated.

Is that correct?

cetmix_tower_server/models/cx_tower_jet_dependency.py
SirPyTech: I double checked and you are correct, sorry for the noise.

Should we also check that both `jet_id` and `jet_depends_on_id` belong to the same template?

And maybe these template-checking constraints might go in another constraint because they are not checking self-dependency after all.

cetmix_tower_server/models/cx_tower_jet.py
SirPyTech: :+1:

cetmix_tower_server/models/res_users.py
SirPyTech: I thought the logic was being introduced here so it could be changed, anyway I see there is already a lot of work behind it even if it's not merged yet.

cetmix_tower_server/models/cx_tower_jet.py
SirPyTech: I would have probably raised an exception here to exit early (without `return`) and highlight the fact that something went wrong; then the caller will choose if the exception should be suppressed.
But it's probably opinion-based because this implementation is ok too.

> This method is mostly used in automation

Note that this `create_waypoint` method is _public_ so it can be called via RPC.

cetmix_tower_server/models/cx_tower_jet_waypoint.py
SirPyTech: :+1:

SirPyTech (commented):
Answered to the open threads of https://github.com/cetmix/cetmix-tower/pull/465#pullrequestreview-3888124918

Aldeigja (approved):
Functional LGTM

cetmix_tower_server/models/cx_tower_command.py
ivs-cetmix: >     * about commands: start/stop are never executed on templates directly.
Technically you can 

> 
>     * about logs: the template field is only used to aggregate the logs of all the jets it has generated.
> 
> 
> Is that correct?

The `jet_template_id` is used to specify for jets of which template the action is triggered.
It will be triggered not for all jets of the same template, but only for the dependent ones.
Eg if the current Odoo jet depends on a Postgress jet but there are some other Postgres jets from the same template, Odoo will be restarted only when the Postgres jet it depends on is restarted. Not any of the Postgres jets in the system.

cetmix_tower_server/models/cetmix_tower.py
ivs-cetmix: It's [there](https://github.com/cetmix/cetmix-tower/blob/ab22f8a1166c29d9ea66db39f26aeedaef4f40c4/cetmix_tower_server/models/cetmix_tower.py#L29), please check again.

cetmix_tower_server/demo/demo_data.xml
ivs-cetmix: Yes, this is on purpose. We are emulating the installation process by simply running an "OK" command.

cetmix_tower_server/models/cetmix_tower.py
SirPyTech: I was opening the file from this diff and landed in an old revision apparently: https://github.com/cetmix/cetmix-tower/pull/465/changes/BASE..99aa8c621c3903f2f4d98a5a7d8766f87c521a0e#diff-23e3084a99c695d248d271b8fdb563c6f702d10d141be3a88ff0f582c5c47b18

Removing the weird `BASE` reference shows the actual changes, and the warning log is there, thanks! :pray:

cetmix_tower_server/models/cx_tower_jet.py
ivs-cetmix: Noted, thank you! This is fixed.

ivs-cetmix:
@SirPyTech thank you for the review! The code was updated, could you please check again?

cetmix_tower_server/models/cx_tower_jet.py
```diff
@@ -0,0 +1,1624 @@
...
+
+        # If the flight plan is finished with an error,
+        # we bring the jet to the error state if it is defined
+        # or back to the initial state if not
+        # Reset the target state because we cannot continue the chain of actions
+        else:
+            vals.update(
+                {
+                    "state_id": (
+                        self.current_action_id.state_error_id
```
alexey-pelykh, line 1183: When the "create" action's flight plan fails: `state_error_id` is not set (optional) → first branch is `False`. `state_from_id` is `False` by definition for create actions → second branch is `False`. Result: `state_id` is set to `False`, leaving the jet stranded with no state.

Consider adding a third fallback to the transit state, or making `state_error_id` required for actions with no `state_from_id`:

```python
"state_id": (
    self.current_action_id.state_error_id
    and self.current_action_id.state_error_id.id
) or (
    self.current_action_id.state_from_id
    and self.current_action_id.state_from_id.id
) or (
    # Fallback: stay in transit state rather than becoming stateless
    self.current_action_id.state_transit_id
    and self.current_action_id.state_transit_id.id
),
```

cetmix_tower_server/models/cx_tower_jet.py
```diff
@@ -0,0 +1,1624 @@
...
+    def _compute_server_allowed_domain(self):
+        """Compute the domain of the server allowed"""
+        for jet in self:
+            jet.server_allowed_domain = (
+                [("id", "in", jet.jet_template_id.server_ids.ids)]
+                if jet.jet_template_id and jet.jet_template_id.server_ids
+                else []
+            )
+
+    @api.depends("jet_template_id", "jet_template_id.action_ids")
```
alexey-pelykh, line 270: This method uses `_get_user_effective_access_level()` to filter `state_available_ids` (line ~297), which depends on the current user's groups and manager role on the jet. Consider adding `@api.depends_context("uid")` alongside the existing `@api.depends` to ensure the ORM properly invalidates the cache across different user sessions:

```python
@api.depends("jet_template_id", "jet_template_id.action_ids")
@api.depends_context("uid")
def _compute_state_available_ids(self):
```

(Note: `_compute_show_available_states` at line 371 already uses `@api.depends_context("uid")` for a similar reason.)

cetmix_tower_server/models/cx_tower_jet.py
```diff
@@ -0,0 +1,1624 @@
...
+        # 1. Requests where the jet is requested explicitly
+        explicit_requests = jet_request_obj.search(
+            [
+                ("jet_id", "=", self.id),  # pylint: disable=no-member
+                ("state", "=", "new"),
+            ]
+        )
+        if explicit_requests:
+            # Check which state is required by the request
+            # TODO: IMPORTANT: we must find a workaround to avoid infinite loops
```
alexey-pelykh, line 1470: The recursive chain `_on_is_available` → `_bring_to_state` → `_finalize_transition` → `_on_is_available` could cause infinite oscillation (or stack overflow) when two jets keep requesting the same target jet in different states.

One possible mitigation: track processed request IDs in context to break the cycle:

```python
processed = set(self.env.context.get('_processed_jet_requests', []))
remaining_requests = remaining_requests.filtered(lambda r: r.id not in processed)
if remaining_requests:
    self = self.with_context(_processed_jet_requests=processed | {remaining_requests[0].id})
    self._bring_to_state(remaining_requests[0].state_requested_id)
```

alexey-pelykh (commented):
Overall this is a well-structured, ambitious feature introducing a comprehensive lifecycle management system for Jet components. The state machine design (actions, transitions, flight plans) is thoughtful, and test coverage is substantial. CI is all green.

A few items I noticed that weren't covered by existing reviews:

**1. Jet can become stateless on initial action failure** (`_flight_plan_finished`)
When the "create" action's flight plan fails, if `state_error_id` is not configured (it's optional), the fallback to `state_from_id` also yields `False` (create actions have no source state by definition). The jet ends up with `state_id = False` — stranded with no state and no way to recover through the normal action path.

**2. `_compute_state_available_ids` filters by user access level but doesn't declare `@api.depends_context("uid")`**
The method calls `_get_user_effective_access_level()` to filter available states, but the decorator doesn't reflect this user dependency. Within a single transaction, different users could see stale cached results from another user's access level.

**3. `_on_is_available` has an acknowledged infinite loop risk with no mitigation**
The TODO correctly identifies that competing jet requests (two jets requesting the same target jet in different states) could cause infinite state oscillation via `_on_is_available` → `_bring_to_state` → `_finalize_transition` → `_on_is_available`. Consider adding a recursion guard or request deduplication before this goes to production.

Co-Reviewed-By: Claude Opus 4.6 <noreply@anthropic.com>

cetmix_tower_server/models/cx_tower_jet.py
ivs-cetmix: That's done on purpose.  If the initial action fails jet must remain stateless.
Also a Jet must not remain in a transit state as that state means that jet is performing some action and is busy.

cetmix_tower_server/models/cx_tower_jet.py
ivs-cetmix: Those are computed non-store fields, their values persist within the current env only.
Note: `_compute_show_available_states` depends on the `uid` to ensure it's triggered to fetch the values store in the current user record.

cetmix_tower_server/models/cx_tower_jet.py
ivs-cetmix: Thank you for the idea, this is actually one of the options we consider.

cetmix_tower_server/demo/demo_data.xml
SirPyTech: :+1:

cetmix_tower_server/models/cx_tower_command.py
SirPyTech: Thanks for the explanation, I can understand the idea behind it and it looks solid :muscle:

> >     * about commands: start/stop are never executed on templates directly.
> 
> Technically you can

Double checking the command model, I see there is a "servers" field so probably start/stopping a jet template would trigger the stop of all the jets created by that template on all the selected servers.
Maybe it would be more clear to change the `help` to something like:

```suggestion
        help="Action will be triggered for all the jets created by this template on the selected servers",
```
The _dependency_ mention is probably what confused me in the first place.

---
> Eg if the current Odoo jet depends on a Postgress jet but there are some other Postgres jets from the same template, Odoo will be restarted only when the Postgres jet it depends on is restarted. Not any of the Postgres jets in the system.

That makes sense, but I think in this case the command could simply be: "stop this Postgres jet", without specifying any template.
Then the command automatically finds out that that specific Postgres jet belongs to an Odoo template, and then stops the depending jets (like the Odoo jet) that are part of that template.

This is why I think it is weird to have an editable `template` field in a command, at least in this example where you just want to stop a jet: because I think the template could be automatically computed.

cetmix_tower_server/models/cx_tower_jet.py
SirPyTech: Thanks for considering this suggestion, now I see an exception is raised :+1:

I'm sorry I did not explicitly mention this earlier but the same reasoning holds for the other early exits that just log an `error` and `return`; I suggest to be consistent throughout this method:
- either always raise an Exception when there is a problem
- or always just log an `error` and return `False`

and please remember to align the method's docstring.

SirPyTech (commented):
Answered to the open threads of https://github.com/cetmix/cetmix-tower/pull/465#pullrequestreview-3888124918

cetmix_tower_server/models/cx_tower_command.py
ivs-cetmix: >That makes sense, but I think in this case the command could simply be: "stop this Postgres jet", without specifying any template.

We cannot depend on jets because commands are server/jet agnostic. Consider this as a broadcast message: "If there is anyone listening to the  Odoo channel - restart".  We cannot be sure if there is anyone at this specific moment in this specific channel. So we are doing a general announcement instead of calling some concrete recipients. And the jet template is that broadcasting channel.

ivs-cetmix: >Double checking the command model, I see there is a "servers" field so probably start/stopping a jet template would trigger the stop of all the jets created by that template on all the selected servers.

You are right, however this is a current technical constraint that jets belong to some concrete server. The MVP implementation has it, however the solution design allows jets from different servers depend on each other. E.g. we may have a Postgres jet on Server A which serves Odoo jets running on other servers.

ivs-cetmix:
@SirPyTech thank you for the review! Could you please check the updates? Would be great to have it merged this week 😄

CetmixGitDrone:
On my way to merge this fine PR!
Prepared branch [16.0-ocabot-merge-pr-465-by-ivs-cetmix-bump-major](https://github.com/cetmix/cetmix-tower/commits/16.0-ocabot-merge-pr-465-by-ivs-cetmix-bump-major), awaiting test results.

CetmixGitDrone:
Congratulations, your PR was merged at ac39147e45fdad89ee949b153831d311d8a57446. Thanks a lot for contributing to cetmix. ❤️

PR #495, opened 2026-09-11
[18.0][ADD] cetmix_tower_ssh_proxy: Connect to Tower servers through SSH jump/proxy servers

GabbasovDinar:
## Summary

Add SSH proxy / jump host support for Cetmix Tower.

The module allows Tower to connect to servers that are not directly reachable from the Tower host by using another Tower server as an SSH gateway.

```text"
Tower
  → Proxy / Bastion Server
    → Target Server
```

The connection is implemented using Paramiko SSH `direct-tcpip` channels. The proxy acts only as a transport layer, while authentication and host key verification for the target server remain handled by Tower.

## Use Cases

This is useful for infrastructures where target servers are:

* available only through a corporate VPN;
* located in private networks;
* accessible through a bastion host;
* using overlapping private IP ranges across different customers.

Example:

```text"
Tower
  → Customer VPN Gateway
    → Corporate VPN
      → Odoo Server
```

The VPN itself is not managed by the module. The proxy server only needs to have network connectivity to the target server.

## Summary by CodeRabbit

* **New Features**
  * Added SSH proxy support for connecting to servers through jump or bastion hosts.
  * Supports multi-hop routes of up to eight servers, including overlapping private networks.
  * Added proxy route configuration for servers, templates, and server creation workflows.
  * Commands, SFTP, and file operations work through proxy connections.
  * Added validation for invalid, circular, archived, or overly long proxy routes.
  * Supports independent host key verification for each connection in a proxy route.

* **Documentation**
  * Added setup, configuration, and usage guidance for SSH proxy connections.


cetmix_tower_ssh_proxy/ssh/proxy.py
```diff
@@ -0,0 +1,246 @@
...
+            raise ValueError(f"Unsupported connection mode: {self.mode}")
+
+        try:
+            ssh_client.connect(**connect_params)
+        except Exception:
+            ssh_client.close()
+            raise
+
+        self._ssh_client = ssh_client
+        return self._ssh_client
```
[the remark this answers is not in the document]
GabbasovDinar: Fixed. Failed target handshakes now close the proxy channel, clean up the full route, and evict the cached manager before the error is propagated.

cetmix_tower_ssh_proxy/ssh/proxy.py
```diff
@@ -0,0 +1,258 @@
...
+        Disconnect SFTP, target and proxy connections and clear the cache.
+        """
+        try:
+            if self.sftp_service._sftp_client is not None:
+                self.sftp_service.disconnect()
+        finally:
+            try:
+                self._disconnect_connections()
+            finally:
+                self.delete_cache(self._cache_key)
```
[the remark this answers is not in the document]
GabbasovDinar: Thanks. We are intentionally keeping the cache process-local and aligned with the existing `SSHManager` implementation in Cetmix Tower.

Odoo worker processes do not share Python memory, so `_connection_cache` is isolated per process. We also don't currently have a reproducible execution path where the same proxy manager is concurrently reused and disconnected from multiple threads within one worker.

Adding per-route synchronization would significantly increase the complexity of this implementation without a demonstrated issue in the current Tower execution model, so I would prefer to keep the existing lifecycle consistent with the upstream SSH manager

cetmix_tower_ssh_proxy/models/cx_tower_server.py
```diff
@@ -0,0 +1,198 @@
...
+        tracking=True,
+        groups="cetmix_tower_server.group_manager",
+        help=(
+            "Jump/bastion server used to reach this server. The proxy may itself "
+            "use another proxy, allowing a multi-hop SSH chain."
+        ),
+    )
+
+    @api.onchange("ssh_connection_route")
+    def _onchange_ssh_connection_route(self):
```
ivs-cetmix, line 41: I would suggest to make the ssh_proxy_server_id computed stored field with readonly=False and replace the onchange with compute.
Because onchange is covering only the changes done via UI, and for example around 80% of all changes we are currently doing in Tower are done via API directly in models.

ivs-cetmix (changes requested):

cetmix_tower_ssh_proxy/tests/test_ssh_proxy.py
ivs-cetmix, line 1: add a regression test, for example before test_circular_route_is_rejected:
```python

    def test_direct_route_clears_proxy_server(self):
        """
        Switching to the Direct route clears SSH Proxy Server outside the UI
        """
        self.target.ssh_connection_route = "direct"
        self.assertFalse(self.target.ssh_proxy_server_id)

        template = self.ServerTemplate.create(
            {
                "name": "Proxy Template",
                "ssh_connection_route": "proxy",
                "ssh_proxy_server_id": self.proxy_a.id,
            }
        )
        template.ssh_connection_route = "direct"
        self.assertFalse(template.ssh_proxy_server_id)

        # Explicit Direct route with a proxy is still rejected
        with self.assertRaises(ValidationError):
            self.proxy_a.write(
                {
                    "ssh_connection_route": "direct",
                    "ssh_proxy_server_id": self.proxy_b.id,
                }
            )

```

ivs-cetmix (changes requested):
@GabbasovDinar thank you for your contribution! Overall LGTM, I would only ask you to add one test to ensure that nothing is broken during the forecoming refactoring:

cetmix_tower_ssh_proxy/models/cx_tower_server.py
GabbasovDinar: fixed

cetmix_tower_ssh_proxy/tests/test_ssh_proxy.py
GabbasovDinar: fixed

cetmix_tower_ssh_proxy/models/cx_tower_server.py
```diff
@@ -0,0 +1,187 @@
...
+
+    @api.depends("ssh_connection_route")
+    def _compute_ssh_proxy_server_id(self):
+        """Clear the SSH proxy server when using a direct route."""
+        for server in self:
+            if server.ssh_connection_route == "direct":
+                server.ssh_proxy_server_id = False
+
+    @api.onchange("ssh_proxy_server_id")
+    def _onchange_ssh_proxy_server_id(self):
```
ivs-cetmix, line 51: Do we still need this function?

ivs-cetmix (changes requested):

